PR #1, opened 2024-01-10
Update docs for manual install on no gpu machines.

funkaoshi:
The `init.sh` script will install on machines without a GPU as well, so include those instructions in the manual instructions as well.


aliasaria:
Thanks!

While it is possible to use Transformer Lab API on a non GPU machine, such as a Mac, it's not a great experience right now so we are leaving that out of the official documentation. After we have some time updating the tooling to support Macs, we can update the docs to show support.

PR #2, opened 2024-01-11
Fix bug that breaks starting models on a new experiment

dadmobile:
Check for inferenceParams key on experiment_config object.

### DIFF
--- a/api.py
+++ b/api.py
@@ -162,7 +162,7 @@ async def server_worker_start(model_name: str, model_filename: str | None = None
 
         experiment_config = experiment['config']
         experiment_config = json.loads(experiment_config)
-        if (experiment_config['inferenceParams'] is not None):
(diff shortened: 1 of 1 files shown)


api.py
```diff
@@ -162,7 +162,7 @@ async def server_worker_start(model_name: str, model_filename: str | None = None
 
         experiment_config = experiment['config']
         experiment_config = json.loads(experiment_config)
-        if (experiment_config['inferenceParams'] is not None):
+        if ('inferenceParams' in experiment_config and experiment_config['inferenceParams'] is not None):
```
funkaoshi, line 165: Can be a bit more succinct with `if experiment_config.get('inferenceParams') is not None:`.  (If we have no inferenceParams's because they are an empty string do you still want to run this blocK?)

aliasaria (approved):

PR #6, opened 2024-02-28
Change homefolder for huggingface models if HF_HOME is set

ai-made-approachable:
Added case that HF_HOME has been set, so different huggingface home folder will be used


dadmobile (approved):
Looks good!

PR #10, opened 2024-05-31
Adding wikipedia to the data gallery

safiyamak:


transformerlab/routers/data.py
```diff
@@ -56,15 +56,17 @@ async def dataset_info(dataset_id: str):
...
-
     r = {}
     # This means it is a custom dataset the user uploaded
     if d["location"] == "local":
         dataset = load_dataset(path=dirs.dataset_dir_by_id(dataset_id))
         # print(dataset['train'].features)
         r["features"] = dataset["train"].features
     else:
-        ds_builder = load_dataset_builder(dataset_id)
+        if (dataset_id == "wikipedia"):
```
dadmobile, line 66: I don't think we should hardcode an if statement for wikipedia in the code.  Instead let's add configuration to the data definition in the gallery and then add code to handle passing a configuration around.

dadmobile (changes requested):
Will work through a plan for how to handle passing configuration around on discord.

transformerlab/db.py
```diff
@@ -26,6 +26,15 @@ async def init():
     await db.execute(
         "CREATE TABLE IF NOT EXISTS dataset(id INTEGER PRIMARY KEY, dataset_id UNIQUE, location, description, size)"
     )
+
+    try:
+        await db.execute(
+            """ALTER TABLE dataset ADD COLUMN json_data TEXT DEFAULT '{}'"""
```
dadmobile, line 32: Should be type JSON I think? See the create statement for model.

dadmobile: Not sure if this has follow-on impact to anything but should probably do the same as model for consistency.

transformerlab/routers/data.py
```diff
@@ -53,18 +53,20 @@ async def dataset_gallery() -> Any:
...
-
     r = {}
     # This means it is a custom dataset the user uploaded
     if d["location"] == "local":
         dataset = load_dataset(path=dirs.dataset_dir_by_id(dataset_id))
         # print(dataset['train'].features)
         r["features"] = dataset["train"].features
     else:
-        ds_builder = load_dataset_builder(dataset_id)
+        dataset_config = d.get("json_data").get("dataset_config")
```
dadmobile, line 65: You can probably just do d["json_data"] if you are assuming it is present (or add {} as a second parameter if you want to use this to guard against a missing json_data parameter).
But then what you can do in the second get is make the second parameter None to make it clear that's what you get if dataset_config doesn't exist?

transformerlab/routers/data.py
```diff
@@ -102,7 +104,11 @@ async def dataset_preview(dataset_id: str = Query(description="The ID of the dat
         result['columns'] = dataset["train"][offset:min(
             offset+limit, dataset_len)]
     else:
-        dataset = load_dataset(dataset_id)
+        dataset_config = d.get("json_data").get("dataset_config")
```
dadmobile, line 107: Same as above.

transformerlab/routers/data.py
```diff
@@ -117,19 +123,30 @@ async def dataset_download(dataset_id: str):
     row = await db.get_dataset(dataset_id)
     if row is not None:
         return {"status": "error", "message": f"A dataset with the name {dataset_id} already exists"}
-
+    # Get the dataset from the gallery
+    file_location = os.path.join(
+        dirs.TFL_SOURCE_CODE_DIR, "transformerlab", "galleries", "data-gallery.json")
+    with open(file_location) as f:
+        gallery = json.load(f)
+    # Gauranteed that the dataset exists in the gallery
```
dadmobile, line 131: I think it's possible to call this function with any string via the API endpoint.  So you should probably set a default value for json_data (I assume {}) so this doesn't crash.

transformerlab/db.py
```diff
@@ -153,26 +162,42 @@ async def get_datasets():
     return data
 
 
-async def create_huggingface_dataset(dataset_id, description, size):
-
+async def create_huggingface_dataset(dataset_id, description, size, json_data):
+    try:
```
dadmobile, line 166: Remove this database migration as well. Will be taken care of in init.

transformerlab/db.py
```diff
@@ -153,26 +162,42 @@ async def get_datasets():
...
         """,
-        (dataset_id, "huggingfacehub", description, size),
+        (dataset_id, "huggingfacehub", description, size, json.dumps(json_data)),
     )
     await db.commit()
 
 
 async def create_local_dataset(dataset_id):
-
+    try:
```
dadmobile, line 186: And this one.

dadmobile (approved):

PR #12, opened 2024-06-07
Adding update template for training jobs and removing duplicates in jobs and training endpoints

safiyamak:
Also moving some train template endpoints into jobs


transformerlab/db.py
```diff
@@ -434,19 +434,20 @@ async def job_cancel_in_progress_jobs():
 ###############
 
 
-async def get_training_template(id):
+# async def get_training_template(id):
 
-    cursor = await db.execute("SELECT * FROM training_template WHERE id = ?", (id,))
-    row = await cursor.fetchone()
+#     cursor = await db.execute("SELECT * FROM training_template WHERE id = ?", (id,))
+#     row = await cursor.fetchone()
```
dadmobile, line 440: I'm not sure I understand why you are commenting out this function?

transformerlab/routers/jobs.py
```diff
@@ -102,3 +103,21 @@ async def get_training_job_output(job_id: str):
...
+async def get_training_template(template_id: str):
+    return await db.get_training_template(template_id)
+
+
+@router.put("/template/update")
+async def update_training_template(template_id: str, name: str,
+                                   description: str,
+                                   type: str,
+                                   config: Annotated[str, Body(embed=True)]):
+    configObject = json.loads(config)
```
dadmobile, line 120: Possibly you can send bad JSON so it might be nice to catch JSONDecodeError and return an error message.

transformerlab/routers/train.py
```diff
@@ -74,11 +69,11 @@ async def delete_training_template(template_id: str):
...
 
-@router.get("/job/create")
-async def job_create(template_id: str, description: str, experiment_id, config: str = "{}"):
-    print("template_id: " + template_id)
-    jobid = await db.training_job_create(template_id=template_id, description=description, experiment_id=experiment_id)
-    return jobid
+# @router.get("/job/create")
+# async def job_create(template_id: str, description: str, experiment_id, config: str = "{}"):
+#     print("template_id: " + template_id)
+#     jobid = await db.training_job_create(template_id=template_id, description=description, experiment_id=experiment_id)
```
dadmobile, line 75: I think you are removing because of a duplicate or something.  If that's the case just remove, don't bother commenting out. We have git history if we need to get it back.

dadmobile (changes requested):

transformerlab/db.py
safiyamak: Oversight sorry, meant to comment out training_job_create

dadmobile (approved):
Merging as this works but we should do somethin about that commented out function.

PR #16, opened 2024-06-24
Removing thread isolation and db commit to avoid db locked error

safiyamak:


transformerlab/shared/download_huggingface_model.py
```diff
@@ -201,7 +200,6 @@ def check_disk_size(model_is_downloaded: Event):
         db = sqlite3.connect(f"{WORKSPACE_DIR}/llmlab.sqlite3")
```
dadmobile, line 200: I think you're missing isolation_level=None here?

transformerlab/shared/download_huggingface_model.py
```diff
@@ -243,7 +241,6 @@ def main():
         db.execute(
```
dadmobile, line 241: Also missing isolation level in the preceding connect statement?

dadmobile (changes requested):

dadmobile (approved):

PR #21, opened 2024-11-22
Update setup.sh add rocm support

Notnaton:


aliasaria:
This is a smart idea to install this in the setup.sh for the inference server.

I suppose we could do the same thing but in our full installer for the whole app, not just for this plugin. Perhaps we merge this for now until we have time to rethink it later.

Thanks so much.

aliasaria:
We're planning on implementing AMD support soon but will do it at a lower level in the app. So closing this PR now.

PR #22, opened 2025-01-06
Make the autotrain plugin work on ubuntu

sanjaycal:
Updated to autotrain plugin to actually work on Ubuntu

### DIFF
--- a/transformerlab/plugins/autotrain_sft_trainer/main.py
+++ b/transformerlab/plugins/autotrain_sft_trainer/main.py
@@ -191,12 +191,9 @@ with subprocess.Popen(
 os.system(
     f"rm -rf {project_name}/autotrain_data")
 
-# Move the model to the TransformerLab directory, as well as rename it
+# Move the model to the TransformerLab directory
 os.system(
-    f"mv {project_name} {output_model_name}")
-
-os.system(
-    f"mv {output_model_name} {config['adaptor_output_dir']}/")
+    f"mv {project_name} {config['adaptor_output_dir']}/")
 
 
 print("Finished training.")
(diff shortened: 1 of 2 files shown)


transformerlab/plugins/autotrain_sft_trainer/main.py
```diff
@@ -187,10 +189,14 @@
...
+    f"rm -rf {project_name}/autotrain_data")
+
+# Move the model to the TransformerLab directory, as well as rename it
+os.system(
+    f"mv {project_name} {output_model_name}")
 
-# Move the model to the TransformerLab directory
 os.system(
-    f"mv {output_model_name} {WORKSPACE_DIR}/models/")
+    f"mv {output_model_name} {config['adaptor_output_dir']}/")
```
aliasaria, line 199: aren't you moving it twice here? first on line L196?

aliasaria:
We need to update the version of the plugin if you make a change to it, to get users to update

transformerlab/plugins/autotrain_sft_trainer/main.py
sanjaycal: That first one is to rename the output model, because I was running into errors with it not being name properly before, I can test it without the rename and see if that works

sanjaycal:
The feedback has now been addressed, the plugin version is updated and the renaming move has been removed

transformerlab/plugins/autotrain_sft_trainer/main.py
```diff
@@ -187,10 +189,11 @@
 # Autotrain outputs its data in a directory named <output_model_name>
```
dadmobile, line 189: Wait I'm confused by this.  All this change did was change every instance of output_model_name to project_name. Why?  Does this comment need to be updated?

sanjaycal:
Should be fixed now

PR #23, opened 2025-01-07
Make the default llama_trainer plugin work on ubuntu

sanjaycal:
Made the default llama_trainer plugin actually work on ubuntu by moving the maximum sequence length and packing parameters to the config instead of the trainer. Also switched configs from TrainingArguments to SFTConfig as TrainingArguments doesnt support packing or maximum sequence length. 

Also removed the check for completion in the callback during training as it was prematurely stopping the training. Instead, the training finished when the trainer decides that the training is complete.

### DIFF
--- a/transformerlab/plugins/llama_trainer/main.py
+++ b/transformerlab/plugins/llama_trainer/main.py
@@ -134,7 +134,8 @@ db.execute(
 )
 db.commit()
 
-max_seq_length = config["maximum_sequence_length"]  # max sequence length for model and packing of the dataset
+max_seq_length = int(config["maximum_sequence_length"])  # max sequence length for model and packing of the dataset
+print(max_seq_length)
 
 args = SFTConfig(
     output_dir=output_dir,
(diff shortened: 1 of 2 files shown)


transformerlab/plugins/llama_trainer/main.py
```diff
@@ -173,9 +175,11 @@ def on_step_end(self, args, state, control, **kwargs):
             # Write to jobs table in database, updating the
             # progress column:
             job.update_progress(progress)
-            if job.should_stop:
-                control.should_training_stop = True
-                return control
+
```
dadmobile, line 178: Why was this stopping the job prematurely? I think the function needs this condition or else the cancel button won't work?

dadmobile (changes requested):
Can you check if the cancel button still works?

transformerlab/plugins/llama_trainer/main.py
sanjaycal: turns out this was just me not understanding the code fully(I think the change to the SFTConfig fixed the problem)

sanjaycal:
It has been uncommented

dadmobile (approved):

PR #29, opened 2025-01-21
Add batch size to llama trainer

sanjaycal:


transformerlab/plugins/llama_trainer/index.json
```diff
@@ -23,6 +23,13 @@
             "minimum": 1,
             "maximum": 4096
         },
+        "batch_size": {
+            "title": "Batch Size",
```
dadmobile, line 27: One quick ask...can you add something to parameters_ui to explain what the setting does?  Something that explains the tradeoff (memory vs. num interations or something?)

dadmobile (changes requested):

transformerlab/plugins/llama_trainer/index.json
sanjaycal: now fixed!

dadmobile (approved):

PR #31, opened 2025-01-24
add inference params to worker start

sanjaycal:


api.py
```diff
@@ -184,122 +184,93 @@ def set_worker_process_id(process):
...
+        error = None
+        if inference_params == "":
+            experiment = await db.experiment_get(experiment_id)
+
+            experiment_config = experiment['config']
+            experiment_config = json.loads(experiment_config)
+            if 'inferenceParams' in experiment_config:
+                inference_params = experiment_config['inferenceParams']
+                if inference_params is not None:
+                    inference_params = json.loads(inference_params)
```
dadmobile, line 200: Probably should put in a try/catch.  (or kill the else below and let this fall through to that try/catch)

api.py
```diff
@@ -184,122 +184,93 @@ def set_worker_process_id(process):
...
 
     if (experiment_id is not None):
-        experiment = await db.experiment_get(experiment_id)
-
-        experiment_config = experiment['config']
-        experiment_config = json.loads(experiment_config)
-        if ('inferenceParams' in experiment_config and experiment_config['inferenceParams'] is not None):
-            inference_params = experiment_config['inferenceParams']
+        error = None
+        if inference_params == "":
```
dadmobile, line 192: Throw a comment in here saying something about trying to get params from the experiment if none were passed in?

dadmobile (changes requested):

api.py
sanjaycal: I just reformatted the code to be a lot cleaner(and liberally used try/catch)

api.py
sanjaycal: yup now done!

dadmobile (approved):

PR #33, opened 2025-01-27
Fix issues for MLX LoRA for making model fusing optional as per #42 on app repo

deep1401:


transformerlab/plugins/mlx_lora_trainer/index.json
```diff
@@ -113,6 +122,9 @@
     },
     "steps_per_eval": {
       "ui:help": "Number of training steps between validations."
+    },
+    "fuse_model": {
+      "ui:help": "Fuse model after training. Tick this if you'd like to fuse and create a new model after training."
```
dadmobile, line 127: Can you change this to say something about what the tradeoff here is?  So maybe "Fuse adapter and model together. If unchecked, this will create a separate adapter" or something like that?

transformerlab/plugins/mlx_lora_trainer/main.py
```diff
@@ -221,10 +223,10 @@
 print(example)
 
 # TODO: For now create adapter in the plugin directory but this should probably go somewhere else
-adaptor_output_dir = plugin_dir
-# adaptor_output_dir = config["adaptor_output_dir"]
-# if not os.path.exists(adaptor_output_dir):
-#     os.makedirs(adaptor_output_dir)
+# adaptor_output_dir = plugin_dir
```
dadmobile, line 226: This looks like a change of behaviour. Do we pass this directory? Where does it end up?  If you know maybe change the comment above this section?

dadmobile (changes requested):

transformerlab/plugins/mlx_lora_trainer/main.py
deep1401: @dadmobile 
So adaptor_output_dir is passed as a path ~/.transformerlab/workspace/adaptors/<model name separated by slashes>. For example, I trained 3 adapters and all of them were stored in /Users/deep.gandhi/.transformerlab/workspace/adaptors/google/gemma-2-2b-it as separate folders. 
This behaviour enables loading adapters in a specific model from the Foundation model page. 

Earlier the plugin_dir just led to all files being dumped in `~/.transformerlab/workspace/plugins/mlx_lora_trainer` which seemed counterintuitive as those safetensors weren't being used anywhere

PR #39, opened 2025-01-29
Creating a new plugin for MLX and removing Fastchat dependency for evals

deep1401:


transformerlab/plugins/eleuther-ai-lm-evaluation-harness-mlx/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Clone the repository
+git clone https://github.com/chimezie/lm-evaluation-harness-mlx || { echo "Git clone failed or repository already exists"; exit 1; }
```
dadmobile, line 3: Need to be able to run setup over and over (every time version increments) so I think you have to let this fail or handle the case of the repo being there?

transformerlab/plugins/eleuther-ai-lm-evaluation-harness/main.py
```diff
@@ -40,22 +40,22 @@
...
-
-    # first check http://localhost:8338/healthz
-    response = requests.get('http://localhost:8338/server/worker_healthz')
-    print(response.json())
-    if response.status_code != 200 or not isinstance(response.json(), list) or len(response.json()) == 0:
-        print("Local completions server is not running. Please start it before running the evaluation.")
-        sys.exit(0)
+    # print("CUDA is not available. Running eval using the MLX Plugin.")
+    print("CUDA is not available. Please use the `eleuther-ai-lm-evaluation-harness-mlx-plugin` if using a Mac.")
+    sys.exit(0)
```
dadmobile, line 45: 0 means success?  Maybe it should return 1 or something?

dadmobile (changes requested):
Just small things!

dadmobile (changes requested):
Possibly you also need to increment the version on the non-MLX plugin?

transformerlab/plugins/eleuther-ai-lm-evaluation-harness/main.py
deep1401: Currently if we return sys.exit(1), the pipe which writes the log to the output file won't work. It is designed to see everything as a success only. Once we integrate the loading functionality in, we'll have better logic for this. Until then, we just handle all errors with sys.exit(0)

transformerlab/plugins/eleuther-ai-lm-evaluation-harness-mlx/setup.sh
deep1401: @dadmobile Pushed the changes now

transformerlab/plugins/eleuther-ai-lm-evaluation-harness/main.py
deep1401: Added

aliasaria (approved):

PR #40, opened 2025-01-29
Make the api run through uv

sanjaycal:


requirements-no-gpu-uv.txt
```diff
@@ -286,16 +286,16 @@ tiktoken==0.8.0
     #   fschat
 tokenizers==0.21.0
     # via transformers
-torch==2.6.0+cpu
+torch==2.5.1+cpu
```
deep1401, line 289: @sanjaycal +cpu is not required here. It works without that

PR #53, opened 2025-02-14
Stop Job Attribute --> Method

abhimazu:
Converted the should_stop attribute from Job class to a method that checks if "stop" is 'true' in the job_data column from Job table.

### DIFF
--- a/transformerlab/plugin_sdk/transformerlab/plugin.py
+++ b/transformerlab/plugin_sdk/transformerlab/plugin.py
@@ -56,8 +56,26 @@ class Job:
 
     def __init__(self, job_id):
         self.id = job_id
-        self.db = get_db_connection()
-        self.should_stop = False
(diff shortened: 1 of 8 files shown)


transformerlab/plugin_sdk/transformerlab/plugin.py
```diff
@@ -56,8 +56,26 @@ class Job:
 
     def __init__(self, job_id):
         self.id = job_id
-        self.db = get_db_connection()
-        self.should_stop = False
+        dbfile = os.path.join(WORKSPACE_DIR, "llmlab.sqlite3")
```
dadmobile, line 59: Will talk to you on discord about the need for this but also, even if we go with this for now, we should think abou thow to minimize redefining the DB location etc. and centralize this sort of thing.  There's an interesting question for how we should do this for the plugin SDK.  
Anyways, let's figure everything else first and then decide what to do with this.

PR #56, opened 2025-02-20
Cancel model download

abhimazu:
1) Users can now cancel a model download using the cancel download button.
2) Auto deletion of partially downloaded files
3) Added a general functionality to kill a process if the 'stop' is set to True.

### DIFF
--- a/transformerlab/db.py
+++ b/transformerlab/db.py
@@ -465,6 +465,22 @@ async def job_cancel_in_progress_jobs():
     return
 
 
+async def job_get_job_data(job_id):
+    cursor = await db.execute("SELECT job_data FROM job WHERE id = ?", (job_id,))
+    row = await cursor.fetchone()
+    await cursor.close()
+
+    if row is None:
+        raise ValueError(f"No job found with id: {job_id}")
+
+    job_data = row[0]
+
+    if isinstance(job_data, str):
+        return json.loads(job_data)
+    else:
+        return job_data
+
+
 async def job_update_job_data_insert_key_value(job_id, key, value):
     value = json.dumps(value)
 
--- a/transformerlab/shared/shared.py
+++ b/transformerlab/shared/shared.py
@@ -7,6 +7,8 @@ import threading
 import re
 import time
 import unicodedata
+from pathlib import Path
+import shutil
 from transformerlab.routers.experiment.evals import run_evaluation_script
 from transformerlab.routers.experiment.generations import run_generation_script
 from transformerlab.shared import dirs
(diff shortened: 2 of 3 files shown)


transformerlab/db.py
```diff
@@ -465,6 +465,22 @@ async def job_cancel_in_progress_jobs():
     return
 
 
+async def job_get_job_data(job_id):
```
dadmobile, line 468: Any reason you didn't want to just use job_get?

transformerlab/db.py
```diff
@@ -465,6 +465,22 @@ async def job_cancel_in_progress_jobs():
...
+
+    if row is None:
+        raise ValueError(f"No job found with id: {job_id}")
+
+    job_data = row[0]
+
+    if isinstance(job_data, str):
+        return json.loads(job_data)
+    else:
+        return job_data
```
dadmobile, line 481: What's this?

transformerlab/shared/shared.py
```diff
@@ -364,3 +381,42 @@ def print_in_rainbow(text):
             print(chunk, end="")
             print(reset, end="")
         print("", flush=True)
+
+
+def delete_model_from_hf_cache(model_id: str, cache_dir: str = None) -> None:
```
dadmobile, line 386: Maybe this should go somewhere else?  models/huggingfacemodel.py maybe? Eventually we'll probably want uninstall methods in there anyways. Or is there a reason it should be in here?

transformerlab/shared/shared.py
```diff
@@ -364,3 +381,42 @@ def print_in_rainbow(text):
...
+    If cache_dir is not provided, it defaults to:
+        - $HF_HOME/hub (if HF_HOME environment variable is set), or
+        - ~/.cache/huggingface/hub
+
+    Args:
+        model_id (str): The model ID to delete from the cache.
+        cache_dir (str, optional): The Hugging Face cache directory. Defaults to None.
+    """
+    # Determine the cache directory
+    if cache_dir is None:
```
dadmobile, line 405: This could also be a seprate function but it doesn't look like we use it anywhere else so I guess don't worry about that for now!

dadmobile (changes requested):
This is more just request some answers than changes. The core code looks good.

transformerlab/db.py
abhimazu: We can use job get as well, but I figured later for other features, we might just want to reuse this to get the job_data dictionary.

transformerlab/shared/shared.py
abhimazu: We can shift it anywhere that you see fit. Since it will also be used by the next feature that enables delete of downloaded models, I kept it in shared.

transformerlab/shared/shared.py
abhimazu: Got it, we can move it to a separate function when needed.

transformerlab/db.py
dadmobile: OK I'm not sure that's possible in sqlite though?

transformerlab/db.py
dadmobile: I think just for the sake of minimizing code we have to maintain, let's jsut use job_get?

transformerlab/shared/shared.py
dadmobile: OK great. For now let's put huggingface related functions in models/huggingfacemodel.py.  I might be reusing some of this code there as well.

dadmobile (approved):

PR #58, opened 2025-02-21
Enable multiple runs to be shown in one tensorboard for the same task

deep1401:


aliasaria (changes requested):
When I run this locally, I get directories that look like:
20250221-124539, 20250221-125018
Which implies that jobId isn't getting added to the directory name, just the date and time

transformerlab/routers/train.py
```diff
@@ -346,10 +346,24 @@ async def spawn_tensorboard(job_id: str):
...
+    experiment_dir = dirs.experiment_dir_by_name(data["name"])
+
+    job_data = job["job_data"]
+
+    if "template_name" not in job_data.keys():
+        raise ValueError("Template Name not found in job data")
+
+    os.makedirs(f"{experiment_dir}/tensorboards/{job_data['template_name']}", exist_ok=True)
 
     # hardcoded for now, later on we should get the information from the job id in SQLITE
```
aliasaria, line 365: we can remove this comment as it no longer applies now -- we are going to the db

aliasaria:
> When I run this locally, I get directories that look like: 20250221-124539, 20250221-125018 Which implies that jobId isn't getting added to the directory name, just the date and time

My mistake -- I forgot to update the plugins!

PR #62, opened 2025-02-21
Update Delete HF Model from Cache

abhimazu:
### Fix: 

- Delete HF Cache 

### Affecting: 

- Huggingface models

### Changes:

- Updated the delete model from cache function to use 'scan cache dir' to get model list and get path from that list instead. 

### Steps to test:

- Cancel download while downloading a model 
- You should not see the model folder inside the cache folder post cancel download


aliasaria:
I can't merge because ruff check is not passing

aliasaria:
Now there is a conflict :)

PR #66, opened 2025-02-24
Remove unneeded requirements and update readme

sanjaycal:


README.md
```diff
@@ -59,20 +59,15 @@ Once conda and dependencies are installed, run the following:
...
-    requirements-gpu.in requirements.in
+uv pip compile requirements.in -o requirements-uv.txt
 
 # requirements for systmes without GPU support
-pip-compile \
-    --extra-index-url=https://download.pytorch.org/whl/cpu \
-    --output-file=requirements-no-gpu.txt \
-    requirements.in
+uv pip compile requirements.in -o requirements-no-gpu-uv.txt --extra-index-url=https://download.pytorch.org/whl/cpu
+sed -i 's/\+cpu//g' requirements-no-gpu-uv.txt
```
aliasaria, line 70: 👀

sanjaycal: I need the sed to get rid of the +cpu s that uv pip compile adds for some reason

sanjaycal: (also I just realized that the pytest assumes there is a requirements.txt available, so I'm going to leave it in as a duplicate of requirements-uv.txt)

aliasaria: Got it -- ok any of these weird things we have to do, please leave a comment in the code so that the next person will know why we had to do it.

sanjaycal: should be done now

aliasaria (approved):

PR #74, opened 2025-02-25
Fix/use secure filename

aliasaria:
use secure_filename() in more places. Lots more to go


deep1401:
@aliasaria I'll look at this and add it to other places too

PR #79, opened 2025-02-26
Tensorboard to evals and improved job data

deep1401:
- Adding tensorboard to evals 
- Improved the job data within evals to match training ones closely
- Add a function to add dynamic json fields to job data

### DIFF
--- a/transformerlab/plugin_sdk/transformerlab/plugin.py
+++ b/transformerlab/plugin_sdk/transformerlab/plugin.py
@@ -135,6 +135,18 @@ class Job:
             (tensorboard_dir, self.id),
         )
 
+    def add_to_job_data(self, key: str, value: str):
+        """
+        Adds a key-value pair to the job_data JSON object.
+        """
+        try:
+            self.db.execute(
+                "UPDATE job SET job_data = json_insert(job_data, '$." + key + "', ?) WHERE id = ?",
+                (value, self.id),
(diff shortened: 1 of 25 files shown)


transformerlab/plugin_sdk/transformerlab/plugin.py
```diff
@@ -135,6 +135,18 @@ def set_tensorboard_output_dir(self, tensorboard_dir: str):
             (tensorboard_dir, self.id),
         )
 
+    def add_to_job_data(self, key: str, value: str):
```
abhimazu, line 138: There is a function inside db.py named "job_update_job_data_insert_key_value" that does this already.

abhimazu (changes requested):
Just the update key value pair function in the db file can be used instead.

transformerlab/plugin_sdk/transformerlab/plugin.py
deep1401: @abhimazu  I cannot access the db.py functions directly within a plugin, adding a function would be cleaner than having all db functions use an added path to access the entire transformerlab code directory rather than just plugin.py

abhimazu: Got it..

abhimazu (approved):

PR #82, opened 2025-02-28
Fixes coloured evals

deep1401:
- Add coloured eval states to the custom plugins developed by us
- Not adding to Eleuther as it internalizes normally and thus, it will automatically be assumed to be within our increasing pattern


aliasaria (changes requested):
Can you explain this to me more. I want to make sure we keep this simple for our users who will generate scores in the future.

PR #86, opened 2025-03-03
Docker support

abhimazu:
#### Fix:
- Dockerfile Update

#### Affecting:
- Container image build and run

#### Changes:
- Added dockerfiles for cpu and gpu along with latest docker-compose file

#### Steps to test:

- Checkout the branch
- Run "docker compose up -d"


Dockerfile.gpu
```diff
@@ -0,0 +1,27 @@
...
+
+SHELL ["/bin/bash", "--login", "-c"]
+
+# Set noninteractive mode
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install minimal dependencies required for the install script
+RUN apt-get update && apt-get install -y \
+    curl \
+    python3.11 \
```
aliasaria, line 11: Why are you installing python when our tool installs python?

aliasaria (approved):

aliasaria:
The reason we might not want to call the file "gpu" is because there are other gpus out there -- we will need a docker file for amd gpus later

PR #87, opened 2025-03-03
Model Provenance

abhimazu:
#### Add:
- Model Provenance

#### Affecting:
- None

#### Changes:
- Added a route to provide model provenance data for a target model

#### Steps to test:

- Checkout the branch and run the 'run.sh'
- Test the route "/model/provenance/{model_id}" by providing a target model id downloaded in your local using swagger UI.


aliasaria (commented):
A model maybe be imported from another user or system. So we shouldn't trace provenance by going through all jobs. Provenance should be stored, not deduced. So that means every time a train or mutating job happens (this includes things like quantization, etc) it should append the data somewhere, instead of trying to "trace back".

But if we want to check in this work to show the UI, that is ok.

aliasaria (approved):

PR #114, opened 2025-03-04
Dataset download for fine-tuning Embedding models

abhimazu:
#### Add:
- Dataset Download for Embedding Models fine-tuning

#### Affecting:
- Data router 

#### Changes:
- Added conditional dataset loader param 'name' for datasets that are needed for embedding model finetuning.

#### Steps to test:

- Checkout the branch
- Use swagger UI to download a new dataset for Embedding model fine-tuning using both the params. The change in UI will follow this PR. 
- After download success from Swagger UI, check if the dataset shows up in your app. Additionally, click on preview and info for more details.


transformerlab/routers/data.py
```diff
@@ -255,38 +264,60 @@ async def dataset_download(dataset_id: str):
...
 
     dataset_size = ds_builder.info.download_size
     if not dataset_size:
         dataset_size = -1
+
+    if json_data == {}:
+        json_data = {
+            "name": ds_builder.info.dataset_name,
+            "huggingfacerepo": dataset_id,
+            "config_name": ds_builder.info.config_name,
```
deep1401, line 288: Is this set to None in normal cases?

deep1401 (approved):
Approving as it doesn't break anything as discussed with Abhijeet

PR #115, opened 2025-03-05
Removing Restricted Logic for LoRA Train Type

deep1401:
- Set a train_type within the index.json, if not specified "LoRA" will be used
- Add logic for running a pre-training run 
- Fix bugs and remove redundant parameters for pre-training
- Convert pre-trained model to huggingface format
- Fix import bugs with locally importing models within Transformer Lab due to the new file system

### DIFF
--- a/transformerlab/models/filesystemmodel.py
+++ b/transformerlab/models/filesystemmodel.py
@@ -61,6 +61,7 @@ class FilesystemModel(basemodel.BaseModel):
         self.source = "local"
         self.source_id_or_path = model_path
         self.model_filename = model_path
+        self.model_path = model_path
 
     async def get_json_data(self):
         json_data = await super().get_json_data()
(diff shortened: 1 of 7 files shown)


aliasaria (approved):

transformerlab/models/filesystemmodel.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, model_path):
         self.source = "local"
         self.source_id_or_path = model_path
         self.model_filename = model_path
+        self.model_path = model_path
```
abhimazu, line 65: All the three vars in line 62, 63 and 64 are set to model_path ? Are they used separately in context ?

transformerlab/plugins/nanotron_pretrainer/convert_nanotron_to_hf.py
```diff
@@ -0,0 +1,155 @@
...
+    # This function selects the proper chunk in the bundled weights to return
+    # either the gate or the up projection only.
+    weight_size = gate_up_proj.shape[0] // 2
+    if gate:
+        return gate_up_proj[:weight_size]
+    else:
+        return gate_up_proj[weight_size:]
+
+
+def convert_nt_to_hf(nanotron_model: LlamaForTraining, hf_model: LlamaForCausalLM, model_config: NanotronLlamaConfig):
```
abhimazu, line 64: Consider moving the model conversion and generic model functions to the 'models' folder if they will be used somewhere in the future, it might help in developing a generic pre-training framework later.

deep1401: This is all nanotron specific only

PR #122, opened 2025-03-06
Create tensorboard directory if it doesn't exist and add missing imports

deep1401:


aliasaria (approved):
Later on, once we clean up plugins, let's remove this from the main.py and put it in the library or somewhere

PR #124, opened 2025-03-07
Fix/Embedding Models

abhimazu:
#### Fix:

- Embedding Model Plugin

#### Changes:

- Created new entry for embedding job_type inside shared.py
- Mapped the entry with index.json

### DIFF
--- a/transformerlab/plugins/embedding_model_trainer/main.py
+++ b/transformerlab/plugins/embedding_model_trainer/main.py
@@ -129,8 +129,8 @@ job_id = config["job_id"]
 
 # Set final model name from model_id, template_name, and job_id
 template_name = config.get("template_name", "Fine-tune-embed-" + time.strftime("%Y%m%d-%H%M%S"))
-model_id = model_id.replace("/", "~~~")
-final_model_name = f"{model_id}_{template_name}_{job_id}"
+model_id_output = model_id.replace("/", "~~~")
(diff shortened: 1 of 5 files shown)


transformerlab/plugins/embedding_model_trainer/main.py
```diff
@@ -227,8 +227,11 @@ def get_loss_function(loss_name, model):
 
 print(f"Loading Sentence Transformer model {model_id}")
 try:
-    model = SentenceTransformer(model_id, device=("cuda" if torch.cuda.is_available() else "cpu"))
+    model = SentenceTransformer(
+        model_id, device=("cuda" if torch.cuda.is_available() else "cpu"), local_files_only=os.path.exists(model_id)
```
deep1401, line 231: Set this using a device_map thing if it exists in SentenceTransformer otherwise, all gpus will always be used. You can then add a parameter which takes in gpu_ids for training this

deep1401 (changes requested):

transformerlab/plugins/embedding_model_trainer/main.py
abhimazu: The device_map is not supported in Sentence Transformers unfortunately. We need to figure that out in next version.

deep1401 (approved):

PR #125, opened 2025-03-07
Fastchat Multi GPU

deep1401:
- Set default as multi-gpu
- Option to set gpu_ids if needed

### DIFF
--- a/transformerlab/plugins/fastchat_server/index.json
+++ b/transformerlab/plugins/fastchat_server/index.json
@@ -36,7 +36,7 @@
   },
   "parameters_ui": {
     "gpu_ids": {
-      "ui:help": "Used to spread models over multiple GPUs. Specify a comma-separated list of GPU IDs to use for inference. For example: 0,1,2,3"
+      "ui:help": "Specify a comma-separated list of GPU IDs to use for inference. The IDs for each GPU can be found in the Computer tab. For example: 0,1,2,3"
     }
   }
 }
--- a/transformerlab/plugins/fastchat_server/index.json
+++ b/transformerlab/plugins/fastchat_server/index.json
@@ -24,9 +24,9 @@
   "setup-script": "setup.sh",
   "parameters": {
   "gpu_ids": {
-          "title": "GPU IDs to use for Inference",
+          "title": "GPU IDs to use for Inference. Leaving blank will use all available GPUs",
           "type": "string",
-          "default": "0"
+          "default": ""
       },
     "eight_bit": {
       "title": "Enable 8-bit compression",
--- a/transformerlab/plugins/fastchat_server/index.json
+++ b/transformerlab/plugins/fastchat_server/index.json
@@ -22,10 +22,11 @@
   "files": ["main.py", "setup.sh"],
   "setup-script": "setup.sh",
   "parameters": {
-    "num_gpus": {
-      "title": "Number of GPUs",
-      "type": "integer"
-    },
+  "gpu_ids": {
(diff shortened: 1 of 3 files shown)


transformerlab/plugins/fastchat_server/index.json
```diff
@@ -22,20 +22,20 @@
   "files": ["main.py", "setup.sh"],
   "setup-script": "setup.sh",
   "parameters": {
-    "num_gpus": {
-      "title": "Number of GPUs",
-      "type": "integer"
-    },
+  "gpu_ids": {
+          "title": "GPU IDs to use for Inference",
```
dadmobile, line 26: This is a little weird to put in the UI.  How will the user know what numbers to put? 
I guess we could list as an "advanced" parameter but then we should:
- make it clear what the default behaviour is (i.e. that it will use all of your GPUs)
- tell the user they can find the IDs on the settings page? (that's true, right?)

transformerlab/plugins/fastchat_server/main.py
```diff
@@ -31,8 +31,16 @@
 else:
     eight_bit = False
 
-# Used only if memory on multiple cards is needed. Do not include if this is not not set.
-num_gpus = parameters.get("num_gpus", None)
+gpu_ids = parameters.get("gpu_ids", "")
+if gpu_ids is not None and gpu_ids != "":
+    num_gpus = len(gpu_ids.split(","))
```
dadmobile, line 36: What happens if the GPU ID list isn't formatted correctly? Do we just pass that on to fastchat? What does it do?

transformerlab/plugins/fastchat_server/index.json
```diff
@@ -22,20 +22,20 @@
...
+          "type": "string",
+          "default": "0"
```
dadmobile, line 28: Oh wait so default is to only use the first GPU? Then we have to say that and then explain to use all GPUs you pass empty string?  IS that right?

dadmobile (changes requested):
I don't quite understand how to use it. Leaving a few comments.

transformerlab/plugins/fastchat_server/main.py
deep1401: Corrected this to check if formatting is not proper, it will just default to use all GPUs if the formatting is incorrect

transformerlab/plugins/fastchat_server/index.json
deep1401: No this was a bug, I have pushed a fix, default is to use all GPUs

transformerlab/plugins/fastchat_server/main.py
```diff
@@ -31,8 +35,22 @@
...
 
-# Used only if memory on multiple cards is needed. Do not include if this is not not set.
-num_gpus = parameters.get("num_gpus", None)
+gpu_ids = parameters.get("gpu_ids", "")
+if gpu_ids is not None and gpu_ids != "":
+    gpu_ids_formatted = gpu_ids.split(",")
+    if len(gpu_ids_formatted) > 1:
+        num_gpus = len(gpu_ids_formatted)
+        # To remove any spacing issues which may arise
+        gpu_ids = ",".join([gpu_id.strip() for gpu_id in gpu_ids_formatted])
```
dadmobile, line 44: OK I see we are checking for number if there's one GPU parameter.  I think you can still mess this up with a poorly formatted string but let's just say this is an advanced parameter and accept it for now.

dadmobile (approved):

PR #127, opened 2025-03-07
Fix completions when fields provided by token data are different

deep1401:


transformerlab/fastchat_openai_api.py
```diff
@@ -670,20 +670,34 @@ def convert_to_openai_format(token_data):
         print("Token Data is not a dictionary, returning None")
         return None
 
+    if 'token_logprobs' in token_data:
```
dadmobile, line 673: This is because fastchat returns an array instead of a single token sometimes?  Can you maybe explain in the docs above? 
i.e. update the part in the pydocs where it shows the Input format to show what the alternate format is and explain when that happens?

dadmobile (changes requested):

transformerlab/fastchat_openai_api.py
deep1401: @dadmobile Added now

dadmobile (approved):

PR #132, opened 2025-03-10
Fix to set fallback system prompt for GRPO

deep1401:


transformerlab/plugins/grpo_trainer_multi_gpu/main.py
```diff
@@ -372,7 +384,7 @@ def soft_format_reward_func(completions, **kwargs) -> list[float]:
     gradient_accumulation_steps=2,
     gradient_checkpointing=True,
     optim="paged_adamw_32bit",
-    logging_steps=10,
+    logging_steps=1,
```
dadmobile, line 394: Just verifying this was an intentional change?

deep1401: I find 1 to be better but I reverted for it to be uniform to everything else.

aliasaria (approved):

PR #134, opened 2025-03-10
Version upgrade for wandb

deep1401:
- Fixes protobuf degradation errors caused in some devices


aliasaria (changes requested):
No need to add requirements-gpu if you are adding it to requirements-no-gpu-uv.txt

deep1401:
> No need to add requirements-gpu if you are adding it to requirements-no-gpu-uv.txt

Removed it from there now

aliasaria (approved):

PR #135, opened 2025-03-11
Fix wandb import issues due to protobuf upgrades from mlx-lm

deep1401:
and remove --upgrade flag

### DIFF
--- a/transformerlab/plugins/mlx_lora_trainer/main.py
+++ b/transformerlab/plugins/mlx_lora_trainer/main.py
@@ -68,11 +68,18 @@ import re
 import subprocess
 import sys
 import time
-import wandb
 from datasets import load_dataset, get_dataset_split_names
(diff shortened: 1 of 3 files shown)


aliasaria (approved):

transformerlab/plugins/mlx_lora_trainer/main.py
```diff
@@ -68,12 +68,19 @@
...
-import wandb
 from datasets import load_dataset, get_dataset_split_names
 import argparse
 import os
 from tensorboardX import SummaryWriter
 
+try:
+    import wandb
+
+    LOG_WANDB = True
```
dadmobile, line 79: It'd be cool to name this something like WANDB_IMPORTED or something that isn't almost exactly the same as WANDB_LOGGING. :)

PR #136, opened 2025-03-12
Add/plugin/embed dataset gen

abhimazu:
#### Add:

- Embedding model Dataset Generation

#### Affecting:

- Generate data from Docs plugin

#### Changes made:

- Added additional params in plugin config to enable embedding dataset creation
- Added a function to create embedding model datasets for 5 different types based on user provided configs
- Updated the info.md to include steps to run the plugin

#### Steps to test:

- Install the plugin (Reinstall if already installed)
- Check the "Generate Dataset for Embedding Model" option in plugin config
- Select one dataset type from the dropdown underneath the checkbox. Default is "anchor | positive | negative" which is the most common and provides good performance improvements.


transformerlab/plugins/synthesizer_docs/index.json
```diff
@@ -79,6 +96,16 @@
         },
         "max_goldens_per_context": {
             "ui:help": "Specify the maximum number of data points to be generated per context."
+        },
+        "generate_dataset_for_embedding_model": {
+            "ui:help": "Select this option to generate a dataset for fine-tuning embedding models."
+        },
+        "embedding_dataset_type": {
```
dadmobile, line 103: I know we can't hide this so maybe just put "(ignored if the above parameter is not checked)" or something?

dadmobile (changes requested):
Tiny change. Otherwise I will approve!

transformerlab/plugins/synthesizer_docs/index.json
abhimazu: Got it, I had written this in the info.md but I will add here as well.

dadmobile (approved):

PR #142, opened 2025-03-17
Add/docker release build

abhimazu:
#### Add:

- Docker Release Build

#### Affecting:

- NA

#### Changes Made:

- Added a release workflow file for building docker images and pushing to dockerhub on release

#### Steps to test:

- On release, the file must automatically build the latest version, tag images for cpu, gpu and push them to Dockerhub


dadmobile (commented):
I wonder if we should name the file docker.yml instead of release.yml just to be more specific?

PR #143, opened 2025-03-17
Fix deleting adaptors

deep1401:
- Add secure_filename for deleting adaptors

### DIFF
--- a/transformerlab/routers/model.py
+++ b/transformerlab/routers/model.py
@@ -597,6 +597,7 @@ async def model_gets_pefts(
 @router.get("/model/delete_peft")
 async def model_delete_peft(model_id: str, peft: str):
     workspace_dir = dirs.WORKSPACE_DIR
+    model_id = secure_filename(model_id)
     adaptors_dir = f"{workspace_dir}/adaptors/{model_id}"
     peft_path = f"{adaptors_dir}/{peft}"
     shutil.rmtree(peft_path)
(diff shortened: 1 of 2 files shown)


transformerlab/routers/model.py
```diff
@@ -597,9 +597,18 @@ async def model_gets_pefts(
 @router.get("/model/delete_peft")
 async def model_delete_peft(model_id: str, peft: str):
     workspace_dir = dirs.WORKSPACE_DIR
+    secure_model_id = secure_filename(model_id)
     adaptors_dir = f"{workspace_dir}/adaptors/{model_id}"
     peft_path = f"{adaptors_dir}/{peft}"
```
dadmobile, line 602: Don't need all three `peft_path =` lines.  It will get overwritten by if clause either way.

transformerlab/routers/model.py
```diff
@@ -597,9 +597,18 @@ async def model_gets_pefts(
...
+    # Check if the peft exists
+    if os.path.exists(adaptors_dir):
```
dadmobile, line 604: Is there a reason to do the check? Can we just always use the secure_filename?

dadmobile: (maybe there's a way to create unsecurely or something so you need to check for both? If so can you put a comment and maybe check for secure_filename version first?)

dadmobile (changes requested):

transformerlab/routers/model.py
deep1401: We can directly use the secure_filename, I just put the check for the adapters which may exist from before we implemented that update for secure_filename. We can phase this out after a couple of releases maybe to only use secure_filename? 

I've moved the secure_filename check to happen first

transformerlab/routers/model.py
deep1401: Removed now

dadmobile (approved):

PR #144, opened 2025-03-17
Add/embedding model config

abhimazu:
#### Add:

- Embedding Model first class citizen

#### Affecting:

- Foundation and RAG

#### Changes Made:

- Added config in list models route to send either LLMs or Embedding models list
- RAG will automatically use the loaded model for indexing and query
- Updated llamaindex RAG plugin to use embedding model instead
- Updated RAG route to include embedding model
- Updated list of model architectures for embedding model plugin
- Add reranker to RAG plugin
- Make the generate embeddings use the embedding model


aliasaria:
Let's discuss other ways to accomplish filtering embedding models before merging this

aliasaria (approved):

PR #149, opened 2025-03-19
Add/move more tables to sqlalchemy

aliasaria:
This change moves all table creation over to SQLAlchemy, except for workflows which is WIP.

Tests are all passing but it is a big change.


dadmobile (approved):
Out of scope, but I just wanted to note that we allow a lot of seemingly important stuff to be nullable. :D

PR #153, opened 2025-03-20
Add Batched Generation for RAG based on a dataset

deep1401:


transformerlab/plugins/deepeval_llm_judge/main.py
```diff
@@ -418,22 +418,40 @@ def run_evaluation():
...
+                            retrieval_context=context,
                         )
                     )
             else:
                 for _, row in df.iterrows():
+                    if isinstance(row["context"], list):
+                        context = row["context"]
+                    elif row["context"].startswith("[") and row["context"].endswith("]"):
+                        try:
+                            context = eval(row["context"])
```
dadmobile, line 447: I'm curious what's going on you gotta do this (array is getting JSON stringify-ed in which case maybe you should be doing JSON.parse?). I'm usually scared of adding random evals but I think JSON.parse just called eval anyways. :)

transformerlab/plugins/generate_rag_outputs/main.py
```diff
@@ -0,0 +1,281 @@
...
+from typing import Any, Dict, List
+
+import requests
+from datasets import load_dataset
+
+import transformerlab.plugin
+
+try:
+    # Parse command line arguments
+    parser = argparse.ArgumentParser(description="Run RAG evaluation on dataset")
```
dadmobile, line 17: I take this to mean we haven't added decorators to generate plugins yet? :D

dadmobile (approved):

transformerlab/plugins/deepeval_llm_judge/main.py
deep1401: So DeepEval requires that column in the LLMTestCase object to be a list. This condition just handles things when lets say the list is stringified in the dataset or if it is a numpy array (coming from the RAG batched output plugin) or if it is just a string with no list. So the conditions just handle everything. 
Did the try except on eval just to make sure it is actually a list, if not I just use the raw context directly and pack it in a list. I manually created all these edge cases to make sure nothing breaks lol

transformerlab/plugins/generate_rag_outputs/main.py
deep1401: Not yet :)

PR #159, opened 2025-03-24
Add Code Execution ability in Basic Evals

deep1401:


transformerlab/plugins/basic_evals/info.md
```diff
@@ -61,6 +64,8 @@ Define your own evaluation tasks:
 }
 ```
 
+> Note: Using the `Code` type requires a Python function called `evaluate` that runs the custom code. We only allow the safe, utility and limited builtins from [RestrictedPython](https://github.com/zopefoundation/RestrictedPython). Additionally we also support `re` and `json` modules but no other import is allowed within the code.
```
dadmobile, line 67: Maybe also say something about the return type of evaluate?

dadmobile (changes requested):

transformerlab/plugins/basic_evals/info.md
deep1401: Changed

dadmobile (approved):

PR #172, opened 2025-03-28
Add logic for model activations in Fastchat server

deep1401:


dadmobile (changes requested):
Main comment is just that we should address the CodeQL things.

dadmobile (approved):

PR #173, opened 2025-03-28
Make MLX calculate MLP and Entropy Approximations

deep1401:


transformerlab/plugins/mlx_server/main.py
```diff
@@ -357,6 +359,337 @@ async def api_get_embeddings(request: Request):
 worker = None
 
 
+@app.get("/visualization_available")
+async def check_visualization_available():
```
dadmobile, line 363: Let's go with this for now but we should generalize this to an endpoint that returns a list of capabilities for this server!

PR #177, opened 2025-03-31
Fix training bugs and some other minor bugs

deep1401:
Fixes:
- Fixes tasks queue to remove the second config inside the input['config']
- Fixes Model Provenance to support json coming out of the new db
- Fixes a minor bug which caused issues with saving in embedding-model-trainer
- Fixes a major bug in Embedding Model Trainer which is caused on single GPU machines if using Matryoshka loss and single sentences dataset type

### DIFF
--- a/transformerlab/models/localmodel.py
+++ b/transformerlab/models/localmodel.py
@@ -190,7 +190,13 @@ class LocalModelStore(modelstore.ModelStore):
                 # The configuration is stored as a JSON string inside job_data["config"]
                 config_str = job_data.get("config")
                 try:
-                    config = json.loads(config_str)
+                    if isinstance(config_str, str):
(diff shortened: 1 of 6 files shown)


transformerlab/models/localmodel.py
```diff
@@ -190,7 +190,13 @@ async def build_provenance(self, jobs):
                 # The configuration is stored as a JSON string inside job_data["config"]
                 config_str = job_data.get("config")
                 try:
-                    config = json.loads(config_str)
+                    if isinstance(config_str, str):
```
dadmobile, line 193: Curious what hit this? Possible bug elsewhere in the code.

dadmobile (approved):

PR #179, opened 2025-04-01
Fixes Issue #333 on app

deep1401:
Fix the disk usage on Mac OS by using the diskutil to get the disk usage


aliasaria (commented):
Do we have a generic way to check if the machine is a mac that we can use across the entire API (instead of one off like here)

deep1401:
> Do we have a generic way to check if the machine is a mac that we can use across the entire API (instead of one off like here)

Added a way which uses sys.platform instead

aliasaria (approved):

PR #194, opened 2025-04-07
Add model architecture visualization in Fastchat and MLX

deep1401:
...and change activation endpoint names for Fastchat and MLX

### DIFF
--- a/transformerlab/plugins/fastchat_server/model_worker.py
+++ b/transformerlab/plugins/fastchat_server/model_worker.py
@@ -390,7 +390,7 @@ async def check_visualization_available():
     return {"available": True}
 
 
-@app.post("/worker_generate_with_visualization")
+@app.post("/worker_generate_activation_visualization")
 async def api_generate_with_visualization(request: Request):
     """Generate text with visualization data about activations and attention entropy"""
     params = await request.json()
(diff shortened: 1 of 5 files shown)


transformerlab/plugins/fastchat_server/model_worker.py
```diff
@@ -566,6 +566,67 @@ def compute_attention_entropy(attentions):
...
+@app.post("/worker_generate_layers_visualization")
+async def api_generate_layers_visualization(request: Request):
+    """Generate model architecture visualization data"""
+    try:
+        params = await request.json()
+
+        import numpy as np
+        import re
+
+        def clean_layer_name(layer_name):
```
aliasaria, line 584: Add reference in source code to the opensource project that inspired this code

transformerlab/plugins/fastchat_server/model_worker.py
```diff
@@ -566,6 +566,67 @@ def compute_attention_entropy(attentions):
     return final_np_array
 
 
+@app.get("/architecture_available")
```
aliasaria, line 569: Shouldn't this endpoint name be more specific than "architecture available" for example like supports_visualization or supports_architecture_visualization

aliasaria (changes requested):

aliasaria (approved):

PR #200, opened 2025-04-10
Job SDK for tlab client

deep1401:
Additional things:
- Add model provenance for local models
- Add ability to create provenance files for local models not limited to "Transformerlab/"

### DIFF
--- a/api.py
+++ b/api.py
@@ -45,7 +45,6 @@ from transformerlab.routers import (
     tools,
     batched_prompts,
 )
-from transformerlab.routers.job_sdk import get_xmlrpc_router, get_trainer_xmlrpc_router
 import torch
 from pynvml import nvmlShutdown
 from transformerlab import fastchat_openai_api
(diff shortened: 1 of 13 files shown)


api.py
```diff
@@ -64,6 +63,8 @@
 os.environ["_TFL_WORKSPACE_DIR"] = dirs.WORKSPACE_DIR
 os.environ["_TFL_SOURCE_CODE_DIR"] = dirs.TFL_SOURCE_CODE_DIR
 
+from transformerlab.routers.job_sdk import get_xmlrpc_router, get_trainer_xmlrpc_router
```
aliasaria, line 66: Shouldn't we import the same way we import the other routers above like this:

from transformerlab.routers import (
    data,
    model,
    serverinfo,
    train,
    plugins,
    evals,
    config,
    jobs,
    workflows,
    tasks,
    prompts,
    tools,
    batched_prompts,
)

api.py
```diff
@@ -172,6 +173,7 @@ async def validation_exception_handler(request, exc):
 app.include_router(batched_prompts.router)
 app.include_router(fastchat_openai_api.router)
 app.include_router(get_xmlrpc_router())
+app.include_router(get_trainer_xmlrpc_router())
```
aliasaria, line 176: Oh I see, it's two different routers?

aliasaria (changes requested):

api.py
deep1401: Its a little different than how fastapi registers routers as all the function definitions also need to be registered first before importing

PR #201, opened 2025-04-11
Fix db locks issue when executing remotely

deep1401:
- Adding 5 second wait time for Writer and putting the sync connection in WAL mode
- Consolidating all sync connections to make use of one object only.
- Changed the T5 LoRA trainer to use plugin sdk to remove bare db access

### DIFF
--- a/transformerlab/db.py
+++ b/transformerlab/db.py
@@ -35,9 +35,7 @@ async def init():
     global db
     os.makedirs(os.path.dirname(DATABASE_FILE_NAME), exist_ok=True)
     db = await aiosqlite.connect(DATABASE_FILE_NAME)
-    db.execute("PRAGMA journal_mode=WAL")
-    db.execute("PRAGMA synchronous=normal")
-    db.execute("PRAGMA busy_timeout = 5000")
+
     global db_sync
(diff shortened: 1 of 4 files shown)


transformerlab/db.py
```diff
@@ -34,6 +35,10 @@ async def init():
     global db
     os.makedirs(os.path.dirname(DATABASE_FILE_NAME), exist_ok=True)
     db = await aiosqlite.connect(DATABASE_FILE_NAME)
+    global db_sync
+    db_sync = sqlite3.connect(DATABASE_FILE_NAME, isolation_level=None)
+    db_sync.execute("PRAGMA journal_mode=WAL")
+    db_sync.execute("PRAGMA busy_timeout = 5000")
```
dadmobile, line 41: I think for consistency let's add `PRAGMA synchronous=normal` which we do everywhere else.

dadmobile (changes requested):

transformerlab/db.py
deep1401: Added now and also added PRAGMA for aiosqlite connections

dadmobile (approved):

PR #203, opened 2025-04-14
Change from Miniconda to Miniforge

brunodoamaral:
Miniconda is subject to [Terms of Service from Anaconda](https://legal.anaconda.com/policies/en/), which does not allows usage in companies with more than 200 employees.

https://github.com/transformerlab/transformerlab-api/blob/0c39243151375f934e2caf14e5eea5e23f9dc662/install.sh#L193

As an simple alternative, we could use Miniforge, which AFAIK, has full compatibility with conda commands. The release files can be found here:

https://github.com/conda-forge/miniforge/releases/latest/

As an "upgraded" solution, we could change `conda create` with `mamba create`, which is much faster.

https://github.com/transformerlab/transformerlab-api/blob/0c39243151375f934e2caf14e5eea5e23f9dc662/install.sh#L245

Finally, I'd like to say thanks for bringing this solution as an open source project!

[]'s
Bruno


brunodoamaral:
@aliasaria  as a side note, Issues are not enabled for this repo. So I have to create this PR to bring this issue to your attention.

deep1401:
@brunodoamaral Thanks for your contribution!
We maintain all issues related to transformerlab on the central repository here: https://github.com/transformerlab/transformerlab-app/issues

dadmobile:
Just as a quick update...this looks great, but we're just trying to make sure we have the right migration strategy for users with an existing conda environment.

PR #205, opened 2025-04-14
Install the cloud app, if possible

aliasaria:


dadmobile (commented):
Only comment is that I wonder if the directory should just be called "web" or "app" or something other than static.  But not important!

dadmobile (approved):

aliasaria:
Good point, renamed to webapp

PR #206, opened 2025-04-14
Stop the DB from locking when the file system gets slow

dadmobile:


transformerlab/db.py
deep1401, line 1: Do we need the WAL stuff for the aio connections too?

aliasaria: I think we do

aliasaria (approved):

PR #207, opened 2025-04-14
Fix YourBench Plugin bugs

deep1401:
- Bringing back the venv thing here, as it doesn't work without that on a clean install
- Adding a switch to a specific commit post which YourBench started withdrawing support for Python 3.11


aliasaria (approved):

deep1401:
Going to hold off on this before my uv venv one goes in, closing it currently

PR #210, opened 2025-04-14
Fix DB Sync issues and catch errors

deep1401:


transformerlab/db.py
```diff
@@ -255,29 +250,44 @@ def job_create_sync(type, status, job_data="{}", experiment_id=""):
...
-    cursor = db_sync.cursor()
+    db_sync = None
+    cursor = None
+    try:
+        global DATABASE_FILE_NAME
+        # check if type is allowed
+        if type not in ALLOWED_JOB_TYPES:
+            raise ValueError(f"Job type {type} is not allowed")
+
+        # Use SQLite directly in synchronous mode
```
dadmobile, line 261: Just to be annoying...would it be OK to throw this into a function like "get_db_connection" or something so that the next time we decide to change pragmas we only have to do it in one place?

dadmobile (changes requested):

transformerlab/db.py
deep1401: Fixed now

dadmobile (approved):

PR #211, opened 2025-04-14
Make plugins run with their own venv

deep1401:


dadmobile (commented):
My only comment is that there's a bunch of pretty similar code that checks for venv and builds a generic command to run.  But it doesn't look trivial to factor.

transformerlab/routers/experiment/generations.py
```diff
@@ -246,7 +246,32 @@ async def run_generation_script(experimentId: int, plugin_name: str, generation_
         ]
     )
 
-    subprocess_command = [sys.executable, dirs.PLUGIN_HARNESS] + extra_args
+    # subprocess_command = [sys.executable, dirs.PLUGIN_HARNESS] + extra_args
```
dadmobile, line 249: Might just want to remove this commented out code?

deep1401: Removed now

dadmobile (approved):

PR #220, opened 2025-04-17
Workflow fixes

sanjaycal:


transformerlab/routers/tasks.py
```diff
@@ -191,17 +191,28 @@ async def queue_task(task_id: int, input_override: str = "{}", output_override:s
...
+            job_data["config"]["script_parameters"][key] = inputs[key]
         for key in input_override.keys():
             job_data["config"][key] = input_override[key]
+            job_data["config"]["script_parameters"][key] = input_override[key]
+
+        for key in outputs.keys():
+            job_data["config"][key] = outputs[key]
+            job_data["config"]["script_parameters"][key] = outputs[key]
+        for key in output_override.keys():
+            job_data["config"][key] = output_override[key]
```
dadmobile, line 214: I'm not sure I get how you make sure a key in outputs doesn't override something that is set by inputs?  Do they jsut never have the same name?

transformerlab/routers/workflows.py
```diff
@@ -321,28 +327,55 @@ async def start_next_step_in_workflow():
         if not next_task:
             await db.workflow_run_update_status(workflow_run_id, "FAILED")
             return {"message": f"Could not find task '{task_name}' for workflow node."}
+        
+        previous_node = None
```
dadmobile, line 331: Can you add a comment saying what this is doing?  Or describe before/after?

transformerlab/routers/workflows.py
```diff
@@ -321,28 +327,55 @@ async def start_next_step_in_workflow():
...
+
+        fusePretext = dirs.MODELS_DIR + "/"
 
         if next_task["type"] == "TRAIN":
             next_task["outputs"] = json.loads(next_task["outputs"])
             next_task["outputs"]["adaptor_name"] = str(uuid.uuid4()).replace("-","")
             next_task["outputs"] = json.dumps(next_task["outputs"])
-            if current_job is not None:
-                if current_job["type"] == "GENERATE":
+            if previous_job is not None:
```
dadmobile, line 353: Add a comment here too. A big if matrix is going to be a little wild to maintain. Is there no way to do this dynamically?

dadmobile (changes requested):

transformerlab/routers/workflows.py
dadmobile: Can you add an explanation of what this code is doing?

dadmobile (changes requested):

dadmobile (approved):

PR #225, opened 2025-04-21
Fix pytest issues

deep1401:


transformerlab/plugins/gguf_exporter/index.json
```diff
@@ -18,6 +18,10 @@
         "Phi3ForCausalLM",
         "Qwen2ForCausalLM"
     ],
+    "supported_hardware_architectures": [
+    "cpu",
```
dadmobile, line 22: Actually mlx works here as well (in so much as "mlx" actually just means you are running an apple silicon machine I think?).

dadmobile (changes requested):

transformerlab/plugins/gguf_exporter/index.json
aliasaria: Right, MLX means you have Apple Silicon

deep1401: Changed

dadmobile (approved):

PR #247, opened 2025-04-27
Fix mamba environment creation command

kthakar:
Mamba environment creation path was incorrect and as a result the installation step was failing locally. This fixes the mamba commandline to provide correct path.


aliasaria:
Thank you for pointing this out @kthakar . We issued a hotfix 76df0b7308c4657b8231da0842c6fe26ff703809 that fixes the root issue by using conda (which does have a -k param). But your PR helped us find the root issue. Thank you again!

PR #249, opened 2025-04-28
Add model delete from hf cache logic

deep1401:
- Fixes transformerlab/transformerlab-app#227

### DIFF
--- a/transformerlab/routers/model.py
+++ b/transformerlab/routers/model.py
@@ -572,7 +572,7 @@ async def model_local_create(id: str, name: str, json_data={}):
 
 
 @router.get("/model/delete")
-async def model_local_delete(model_id: str):
+async def model_local_delete(model_id: str, delete_from_cache: bool = False):
     # If this is a locally generated model then actually delete from filesystem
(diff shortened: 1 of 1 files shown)


transformerlab/routers/model.py
```diff
@@ -572,24 +572,36 @@ async def model_local_create(id: str, name: str, json_data={}):
...
         shutil.rmtree(model_path)
 
     else:
-        # If this is a hugging face model then delete from the database but leave in the cache
-        print(
-            f"Deleting model {model_id}. Note that this will not free up space because it remains in the HuggingFace cache."
-        )
-        print("If you want to delete the model from the HuggingFace cache, you must delete it from:")
-        print("~/.cache/huggingface/hub/")
+        if delete_from_cache:
```
dadmobile, line 591: Because of importing there are ways a model could not have an info.json and not be a huggingface model.  But I guess that's fine...it will get caught by the except.  I don't think many people are using import except for huggingface and ollama.

dadmobile (approved):
I think there's a gotcha case but let's merge this as I think it is an edge case that won't get worse!

PR #250, opened 2025-04-28
Fix RAG Reindexing and install the plugin if not already installed

deep1401:
Fixes transformerlab/transformerlab-app#232


dadmobile (changes requested):
It doesn't seem right to be force installing a specific plugin?  Maybe let's chat about this quickly tomorrow so I understand what's going on.

deep1401:
> It doesn't seem right to be force installing a specific plugin? Maybe let's chat about this quickly tomorrow so I understand what's going on.

Resolved now

PR #257, opened 2025-04-29
Add/allow manual install of plugin setup

aliasaria:
Fixes https://github.com/transformerlab/transformerlab-app/issues/192

### DIFF
--- a/transformerlab/routers/plugins.py
+++ b/transformerlab/routers/plugins.py
@@ -91,6 +91,45 @@ async def copy_plugin_files_to_workspace(plugin_id: str):
     copy_tree(plugin_path, dirs.plugin_dir_by_name(plugin_id))
 
 
+async def run_installer_for_plugin(plugin_id: str, log_file):
+    plugin_id = secure_filename(plugin_id)
+    new_directory = os.path.join(dirs.PLUGIN_DIR, plugin_id)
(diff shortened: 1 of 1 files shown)


transformerlab/routers/plugins.py
```diff
@@ -91,6 +91,45 @@ async def copy_plugin_files_to_workspace(plugin_id: str):
...
+    venv_path = os.path.join(new_directory, "venv")
+    plugin_path = os.path.join(dirs.PLUGIN_PRELOADED_GALLERY, plugin_id)
+
+    # Check if plugin exists at the location:
+    if not os.path.exists(plugin_path):
+        print(f"Plugin {plugin_path} not found in gallery.")
+        return {"error": "Plugin not found in gallery."}
+
+    # Open the Plugin index.json:
+    plugin_index_json = open(f"{plugin_path}/index.json", "r")
```
dadmobile, line 106: I think you copied this here from install_plugin but left it in install_plugin just in case?  It is causing a ruff error so I think you can remove it.

dadmobile (changes requested):

transformerlab/routers/plugins.py
aliasaria: Thanks! Fixed

PR #258, opened 2025-04-29
Move inference server model logs to a custom directory

deep1401:
Fixes transformerlab/transformerlab-app#412

### DIFF
--- a/api.py
+++ b/api.py
@@ -182,7 +182,10 @@ worker_process = None
 
 def spawn_fastchat_controller_subprocess():
     global controller_process
-    logfile = open("controller.log", "w")
+    LOGS_DIR = os.path.join(dirs.WORKSPACE_DIR, "logs")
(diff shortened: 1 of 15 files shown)


api.py
```diff
@@ -182,7 +182,10 @@ async def validation_exception_handler(request, exc):
 
 def spawn_fastchat_controller_subprocess():
     global controller_process
-    logfile = open("controller.log", "w")
+    LOGS_DIR = os.path.join(dirs.WORKSPACE_DIR, "logs")
```
aliasaria, line 185: Do this part in dirs.py (joining "logs" to WORKSPACE_DIR and creating it if it doesn't exist

aliasaria (changes requested):

api.py
deep1401: Done now

aliasaria (approved):

PR #263, opened 2025-05-04
Initial changes to support CUDA 12.8 (required for NVIDIA 50-series GPUs)

aliasaria:
In this change we do a few things:
- We add support for CUDA 12.8 in the install.sh script.
- We add support for CUDA 12.8 in the requirements-uv.txt and requirements.in
files using the new `--index` flag. to match the instructions for installing
pytorch according to the pytorch website.
- There is a strange change: we had to stop using $HOST as an env
variable in the run.sh script. This is because it was causing issues with
the conda activate -- somehow something is setting HOST to different value
than the one we are using in the script. So we just change the variable name
- When running plugins, they activate a venv per plugin, this activation
scripts need to know to add the extra index or they won't be able to find
pytorch 2.7 with cuda 12.8
- NOTDONE: We should upgrade all plugin versions to force a rebuild of each venv
- Also note: a clean upgrade of the same conda env didn't work for me,
I had to wipe it and start again. It might work for you.

Fixes: transformerlab/transformerlab-app/issues/422


aliasaria:
Once finalized and merged, this should fix https://github.com/transformerlab/transformerlab-app/issues/422

deep1401:
Additional Notes:
- ROCm testing is completely independent of this upgrade as the wheel is separate for ROCm and testing for that has already been done. 
- Removed the `set -eu` flag to not cause CONDA_BACKUP_CXX errors and also added the appropriate --index flags.
- Made the force reinstall of cuda from conda so it upgrades to 12.8 
- Need to test this update on mac as well. This would require a cpu build for torch I assume

deep1401:
Testing on Mac Notes:
- Works on torch 2.7 after some minor changes to requirements and the pip sync install step.
- A big issue currently present in this is that we treat MacOS as a non-gpu device. Instead, we need to separate out the two so we can install cpu-only wheels for actual non-gpu devices and the torch binary would itself resolve to the mac os arm architecture to not install cuda stuff. Will add that as well to the install as well as setup steps

deep1401:
@aliasaria The ROCm PR has already been tested with torch 2.7 only. Since there are no wheels for torch2.6+rocm anymore, that was a forced upgrade

dadmobile (approved):

PR #264, opened 2025-05-05
Pin DeepEval and add Langchain dependencies wherever necessary

deep1401:


transformerlab/plugins/synthetic_dataset_rag/setup.sh
```diff
@@ -1 +1 @@
-uv pip install langchain langchain-openai anthropic openai datasets PyMuPDF deepeval
\ No newline at end of file
+uv pip install langchain langchain-openai anthropic openai datasets PyMuPDF deepeval==2.8.2
```
aliasaria, line 1: shouldn't we specify version numbers for EVERYTHING?

transformerlab/plugins/synthesizer_scratch/setup.sh
```diff
@@ -1 +1 @@
-uv pip install deepeval langchain-openai chromadb sentence-transformers pypdf instructor anthropic datasets
\ No newline at end of file
+uv pip install deepeval==2.8.2 langchain-openai chromadb sentence-transformers pypdf instructor anthropic datasets langchain langchain-community langchain_text_splitters
```
aliasaria, line 1: Shouldn't we put version number for everything?

aliasaria (changes requested):

transformerlab/plugins/synthesizer_scratch/setup.sh
deep1401: Added now

aliasaria (approved):

PR #268, opened 2025-05-05
Added ability to name exported models for all plugins

aahaanmaini:


transformerlab/plugins/gguf_exporter/main.py
```diff
@@ -25,9 +26,7 @@
 input_model = args.model_name
 outtype = args.outtype
 
-# For internals to work we need the directory and output filename to be the same
-output_filename = args.output_model_id
-output_path = os.path.join(args.output_dir, output_filename)
+output_path = os.path.expanduser(f"~/.transformerlab/workspace/models/{args.exported_model_name}")
```
deep1401, line 29: Could you use the environment variable _TFL_WORKSPACE_DIR here? Or maybe you can import WORKSPACE_DIR from transformerlab.plugin just like get_python_executable instead of hav ing a path defined here

transformerlab/plugins/mlx_exporter/main.py
```diff
@@ -15,9 +15,10 @@
 parser.add_argument("--output_dir", type=str, help="Directory to save the model in.")
 parser.add_argument("--model_name", default="gpt-j-6b", type=str, help="Name of model to export.")
 parser.add_argument("--q_bits", default="4", type=str, help="Bits per weight for quantization.")
+parser.add_argument("--exported_model_name", default="my-exported-model", help="Name of the exported model folder.")
 args, unknown = parser.parse_known_args()
 
-output_path = args.output_dir
+output_path = os.path.expanduser(f"~/.transformerlab/workspace/models/{args.exported_model_name}")
```
deep1401, line 21: Same comment for using WORKSPACE_DIR here

transformerlab/plugins/gguf_exporter/main.py
aahaanmaini: done, thanks!

transformerlab/plugins/mlx_exporter/main.py
aahaanmaini: done, thanks!

deep1401 (commented):
Just tried this one out on the Qwen3-0.6B and it did set the model name correctly but the timestamp was still included in the Model ID part. Is the goal here just to set name or to correct the model_id as well? Just want to clarify this part

dadmobile:
> Just tried this one out on the Qwen3-0.6B and it did set the model name correctly but the timestamp was still included in the Model ID part. Is the goal here just to set name or to correct the model_id as well? Just want to clarify this part

Ideally it would say what the output is going to be (including timestamp) and then you could override that?  

(Although since he's workign on it maybe we should take the opportunity to align the timestamp thing with whatever we do in workflows...do we use job ID?)

aahaanmaini:
Goal was to just update the exported model name, not the model_ID. Would we like to change the model ID as well? We were aiming to make it easier for users to organize their exported models by allowing renaming, but having some additional information (such as timestamp) available in model_ID might be beneficial

transformerlab/plugins/gguf_exporter/main.py
```diff
@@ -16,6 +16,7 @@
 parser.add_argument("--output_dir", type=str, help="Directory to save the model in.")
 parser.add_argument("--output_model_id", type=str, help="Directory to save the model in.")
 parser.add_argument("--model_name", default="gpt-j-6b", type=str, help="Name of model to export.")
+parser.add_argument("--exported_model_name", default="my-exported-model", help="Name of the exported model folder.")
```
dadmobile, line 19: Oh...I think this is actually output_model_id already (two lines above).

dadmobile (commented):
OK interesting...this is slightly more complicated than I thought, but that's OK.  Model anme and output ID are the same thing.  So what I think we need to do is fine the code that generates the proposed output model ID and pull it maybe from the API to the frontend?  And then set that as the default output model name?

deep1401:
> > Just tried this one out on the Qwen3-0.6B and it did set the model name correctly but the timestamp was still included in the Model ID part. Is the goal here just to set name or to correct the model_id as well? Just want to clarify this part
> 
> Ideally it would say what the output is going to be (including timestamp) and then you could override that?
> 
> (Although since he's workign on it maybe we should take the opportunity to align the timestamp thing with whatever we do in workflows...do we use job ID?)

I think we use job_id for those

dadmobile (changes requested):

dadmobile:
I'm going to close this for now.  We can come back to it if we decide to take this approach.

PR #274, opened 2025-05-07
Rewrote exporter plugins to use plugin SDK

aahaanmaini:


transformerlab/plugins/gguf_exporter/main.py
```diff
@@ -1,66 +1,86 @@
 # This plugin exports a model to GGUF format so you can interact and train on a MBP with Apple Silicon
 import os
 import subprocess
-import argparse
 
 from huggingface_hub import snapshot_download
 
 try:
     from transformerlab.plugin import get_python_executable
+    from transformerlab.sdk.v1.export import tlab_exporter
```
dadmobile, line 9: I don't think we need this try block.  I think this was because we changed the format of that directory and we can get rid of this now?  At least I am sure we don't need to add for new plugins!

transformerlab/plugins/gguf_exporter/main.py
```diff
@@ -1,66 +1,86 @@
...
+                "*.py",
+                "tokenizer.model",
+                "*.tiktoken",
+            ],
+        )
+
+    tlab_exporter.progress_update(30)
+    tlab_exporter.add_job_data("status", "Converting model to GGUF format")
+
+    # TODO: This default quantizes to 8-bit. Need to read that in as a parameter.
```
dadmobile, line 57: I don't think this comment is true?  It looks like we are passing in quantization?

dadmobile: (I realize you are just inheriting this but might as well remove. And then we should add updating quantization to your list because GGUF handles a LOT of options It think).

transformerlab/plugins/mlx_exporter/main.py
```diff
@@ -1,47 +1,63 @@
...
+
+@tlab_exporter.exporter_job_wrapper(progress_start=0, progress_end=100)
+def mlx_export():
+    """Export a model to MLX format"""
+
+    # Directory to run conversion subprocess
+    plugin_dir = os.path.realpath(os.path.dirname(__file__))
+    python_executable = get_python_executable(plugin_dir)
+
+    env = os.environ.copy()
```
dadmobile, line 24: Another thing I think you are bringing over from old code...for now it is fine but let's ask the team if we need to do this and if so can we just wrap it somehow so every plugin doesn't do this?

dadmobile (changes requested):

dadmobile (approved):

PR #277, opened 2025-05-09
Change default host to :: for providing support for ipv6 only machine

deep1401:
-- BEFORE MERGING, please check once on WSL.


deep1401:
Closing because using :: as default host requires ipv4 addresses to be mapped to ipv6 and wew should not do this because it would create a lot of issues for existing users and we don't want to ask people to change internal settings on their system for this

PR #285, opened 2025-05-12
Switch from nvidia-ml-py3 to nvidia-ml-py

deep1401:
- Fixes transformerlab/transformerlab-app#448


dadmobile (approved):
Seems to have worked on windows.

PR #287, opened 2025-05-13
Part 1: Added Exporter Logging Functionality

aahaanmaini:
Create output_{job_id}.txt file, matching functionality of training, evaluating and generating and required for the "View Output" button


transformerlab/routers/experiment/export.py
```diff
@@ -116,28 +120,49 @@ async def run_exporter_script(id: int, plugin_name: str, plugin_architecture: st
...
+            )
+            _, stderr = await process.communicate()
+
+            try:
+                stderr_str = stderr.decode("utf-8", errors="replace") 
+            except Exception as e:
+                stderr_str = f"[stderr decode error]: {e}"
+
+            if stderr_str.strip():
+                if "Traceback" in stderr_str or "Error" in stderr_str or "Exception" in stderr_str:
```
dadmobile, line 144: Did you get this from somewhere else?  This feels kind of arbitrary to look for these strings? All just to change the header on the print?

transformerlab/routers/experiment/export.py
```diff
@@ -116,28 +120,49 @@ async def run_exporter_script(id: int, plugin_name: str, plugin_architecture: st
...
+                    f.write(f"\nError output:\n{stderr_str}")
+                else:
+                    print(f"Standard error stream:\n{stderr_str}")
+                    f.write(f"\nStandard error stream:\n{stderr_str}")
+
+            if process.returncode != 0:
+                fail_msg = f"Failed to export model. Return code: {process.returncode}"
+                await db.job_update_status(job_id=job_id, status="FAILED")
+                print(fail_msg)
+                return {"message": fail_msg}
```
dadmobile, line 155: I know it wasn't like this before but can we add "status": "error" to this JSON for consistency with other API calls?

transformerlab/routers/experiment/export.py
```diff
@@ -161,6 +186,7 @@ async def run_exporter_script(id: int, plugin_name: str, plugin_architecture: st
     json.dump(model_description, model_description_file)
     model_description_file.close()
 
+    await db.job_update_status(job_id=job_id, status="COMPLETE")
```
dadmobile, line 189: And then just since we're here and fixing this can you add "status": "success" to the json schema below.

transformerlab/routers/experiment/export.py
```diff
@@ -174,3 +200,61 @@ async def get_export_jobs(id: int):
...
+        # Get job data
+        job = await db.job_get(job_id)
+        job_data = job["job_data"]
+        
+        # Check if it has a custom output file path
+        if job_data.get("output_file_path") is not None:
+            return job_data["output_file_path"]
+        
+        # Get the plugin name from the job data
+        plugin_name = job_data.get("plugin")
```
dadmobile, line 218: We changed this from `exporter_name` to `plugin` which means old jobs won't work but that's fine because this is only used to figure out how to stream output to the console, right?

dadmobile (changes requested):

transformerlab/routers/experiment/export.py
aahaanmaini: Yup sorry, this was something I added to help with debugging, but not necessary now. Removed!

transformerlab/routers/experiment/export.py
```diff
@@ -116,28 +120,49 @@ async def run_exporter_script(id: int, plugin_name: str, plugin_architecture: st
...
+            if stderr_str.strip():
+                if "Traceback" in stderr_str or "Error" in stderr_str or "Exception" in stderr_str:
+                    print(f"Error output: {stderr_str}")
+                    f.write(f"\nError output:\n{stderr_str}")
+                else:
+                    print(f"Standard error stream:\n{stderr_str}")
+                    f.write(f"\nStandard error stream:\n{stderr_str}")
+
+            if process.returncode != 0:
+                fail_msg = f"Failed to export model. Return code: {process.returncode}"
```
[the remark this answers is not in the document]
aahaanmaini: The entire error message does get printed to the terminal and the output file above under with "Error: ...". We can combine them into one output stream if you'd like.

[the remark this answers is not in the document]
aahaanmaini: Updated! Now it returns the error instead of just the code.

transformerlab/routers/experiment/export.py
aahaanmaini: Done!

transformerlab/routers/experiment/export.py
aahaanmaini: Done!

dadmobile (approved):

PR #294, opened 2025-05-14
Add support to run MCP servers in tools tab within Interact

deep1401:


transformerlab/routers/tools.py
```diff
@@ -92,32 +142,142 @@ async def get_prompt():
...
+    if not (ClientSession and StdioServerParameters and stdio_client):
+        return {"status": "error", "message": "MCP client not installed."}
+    params = _get_stdio_server_params(server_file, args=args, env=env)
+    async with stdio_client(params) as (read, write):
+        async with ClientSession(read, write) as session:
+            await session.initialize()
+            return await session.call_tool(tool_id, arguments=arguments or {})
+
+
+@router.get("/install_mcp_server", summary="Install or check MCP server module or script.")
```
dadmobile, line 241: Nitpick but is there any way to output some details to the console here?  I think things are working for me installing servers but there's no output other than looking in console.log in the app to see the API calls are working.

dadmobile (changes requested):

PR #308, opened 2025-05-22
Fix samsum dataset path

mina-parham:


deep1401:
Could you also make this change in transformerlab/galleries?

deep1401 (commented):
Adding this one here as well as I correct the other galleries PR I opened. We need to update size here to 16369

dadmobile (approved):

PR #316, opened 2025-05-26
Fix for Eval and Generate jobs showing status as 'Complete' when stopped

aahaanmaini:


transformerlab/shared/shared.py
```diff
@@ -305,7 +305,11 @@ async def run_job(job_id: str, job_config, experiment_name: str = "default", job
             with open(evals_output_file, "w") as f:
                 f.write("")
         await run_evaluation_script(experiment_id, plugin_name, eval_name, job_id)
-        await db.job_update_status(job_id, "COMPLETE")
+        # Only set to COMPLETE if the job wasn't stopped
+        job_row = await db.job_get(job_id)
+        job_data = job_row.get("job_data", None)
+        if not (job_data and job_data.get("stop", False)):
+            await db.job_update_status(job_id, "COMPLETE")
```
dadmobile, line 312: I think you are saying that if the job was stopped don't change the status and then trust that it was already set to "STOPPED"?  So it would be redundant to set it to "STOPPED" here?  But i can't figure out where that happens (and we don't check for that) so this is kidn of confusing.

dadmobile (changes requested):
Maybe the fix is just a comment or something, or maybe we explicitly set or check the status?

dadmobile (approved):

PR #320, opened 2025-05-28
Add FastAPI Users

dadmobile:


dadmobile:
This is an initial checkin to get FastAPI Users working.  There are some known limitations not covered in this PR that will be fixed before launching:

1. This line in user_service.py
```
SECRET = "TEMPSECRET"
```

2. Using the default JWT implementation of auth just to get going. Not sure if that's the right choice.

requirements-no-gpu-uv.txt
```diff
@@ -308,7 +332,11 @@ pygments==2.19.1
     #   markdown2
     #   rich
 pyjwt==2.10.1
-    # via msal
+    # via
+    #   fastapi-users
+    #   msal
+pyrsmi==0.2.0
```
deep1401, line 338: We only need to have pyrsmi in the requirements-rocm-uv.txt

aliasaria (approved):

PR #322, opened 2025-05-28
Workflow Triggers

aahaanmaini:
Implemented workflow "owned" triggers.


transformerlab/jobs_trigger_processing.py
```diff
@@ -0,0 +1,81 @@
...
+            logger.warning(f"Job {job_id} not found, skipping trigger processing")
+            return
+
+        if job["status"] != "COMPLETE":
+            logger.debug(f"Job {job_id} is not complete (status: {job['status']}), skipping trigger processing")
+            return
+
+        # Step 2: Check if triggers have already been processed
+        job_data = job.get("job_data", {})
+        if job_data.get("triggers_processed", False):
```
aliasaria, line 40: Just wondering why would a jobs triggers already be processed?

transformerlab/db.py
```diff
@@ -29,6 +31,45 @@
 # Create a configured "Session" class
 async_session = sessionmaker(async_engine, expire_on_commit=False, class_=AsyncSession)
 
+# Predefined trigger blueprints - the single source of truth for available trigger types
```
aliasaria, line 34: Do not store this object in db.py. Db.py is a very tightly scoped file only used to store access to the DB but this object is a workflow specific object that should be stored somewhere else and imported in here if needed

transformerlab/db.py
```diff
@@ -444,6 +507,33 @@ def job_update_status_sync(job_id, status, error_msg=None):
...
+            print(f"🎯 Job {job_id} completed, attempting to process triggers")
+            try:
+                from transformerlab.jobs_trigger_processing import process_job_completion_triggers
+                
+                # Since this is a synchronous context, we need to run the async trigger processing
+                # using asyncio. We'll do this in a way that doesn't block if there's already an event loop.
+                
+                # Try to get the current event loop, if it exists
+                try:
+                    import asyncio
```
aliasaria, line 522: do not import inline

transformerlab/db.py
```diff
@@ -45,6 +86,14 @@ async def init():
     async with async_engine.begin() as conn:
         await conn.run_sync(models.Base.metadata.create_all)
 
+    # Add trigger_configs column to workflows if it doesn't exist
```
aliasaria, line 89: Why do we need any of this in the DB? If a workflow is supposed to be triggered by an event, why not just search the list of workflows after every event. That would remove the need for all this stuff in the database entirely.

test/db/test_db.py
```diff
@@ -77,8 +78,31 @@
...
+from sqlalchemy.ext.asyncio import create_async_engine
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.ext.asyncio import AsyncSession
+
+# Create a unique test directory using absolute paths to prevent contamination
+TEST_BASE_DIR = os.path.abspath(os.path.join(tempfile.gettempdir(), f"transformerlab_db_test_{uuid.uuid4().hex[:8]}"))
+os.makedirs(TEST_BASE_DIR, exist_ok=True)
+
+# Set environment variables BEFORE any transformerlab imports
+os.environ["TFL_HOME_DIR"] = TEST_BASE_DIR
```
aliasaria, line 90: You can't move these environment variables settings down here as they are used by the import for db. They have to be set before db is imported.

test/db/test_db.py
```diff
@@ -77,8 +78,31 @@
     delete_training_template,
     export_job_create,
 )
+from sqlalchemy.ext.asyncio import create_async_engine
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.ext.asyncio import AsyncSession
+
+# Create a unique test directory using absolute paths to prevent contamination
+TEST_BASE_DIR = os.path.abspath(os.path.join(tempfile.gettempdir(), f"transformerlab_db_test_{uuid.uuid4().hex[:8]}"))
```
aliasaria, line 86: Don't do this here. If we are going to change the way database files are stored for testing, this should be in a different PR (and something we should discuss separately)

test/db/test_db.py
```diff
@@ -77,8 +78,31 @@
...
+os.makedirs(TEST_BASE_DIR, exist_ok=True)
+
+# Set environment variables BEFORE any transformerlab imports
+os.environ["TFL_HOME_DIR"] = TEST_BASE_DIR
+os.environ["TFL_WORKSPACE_DIR"] = TEST_BASE_DIR
+
+# Patch the database path to ensure complete isolation
```
aliasaria, line 93: Don't change this stuff. This is for a separate PR if we are doing it at all

test/db/test_db.py
```diff
@@ -300,20 +324,18 @@ async def test_export_job_create(test_experiment):
...
+    """Initialize database for testing."""
+    # Database is already patched at import time
     await db.init()
-    yield
-    await db.close()
+    try:
+        yield
+    finally:
+        await db.close()
+        # Clean up test directory
```
aliasaria, line 336: Don't change test functionality in a PR about workflows

test/db/test_db.py
```diff
@@ -682,6 +704,162 @@ async def test_job_update_status_sync(setup_db):
...
+            "trigger_type": bp["trigger_type"],
+            "is_enabled": bp["trigger_type"] == "TRAIN"
+        })
+    await db.workflow_update_trigger_configs(workflow_id, new_configs)
+    
+    # Create a job
+    job_id = await db.job_create("TRAIN", "RUNNING", "{}", test_experiment)
+    
+    # Mock asyncio functions to simulate no running event loop
+    with unittest.mock.patch('asyncio.get_running_loop') as mock_get_loop, \
```
aliasaria, line 726: What exactly are you testing for here?

aliasaria (changes requested):

dadmobile:
I think we either turn this to a draft or close it for now (you can still access it and we can open it later if needed)?

PR #324, opened 2025-05-29
More use of SQLAlchemy

aliasaria:


transformerlab/db.py
```diff
@@ -805,21 +806,20 @@ async def export_job_create(experiment_id, job_data_json):
...
+        return [e.__dict__ for e in experiments]
 
 
 async def experiment_create(name, config):
-    # use python insert and commit command
-    row = await db.execute_insert("INSERT INTO experiment(name, config) VALUES (?, ?)", (name, config))
-    await db.commit()
-    return row[0]
+    async with async_session() as session:
+        experiment = models.Experiment(name=name, config=config)
```
dadmobile, line 818: I'm not exactly sure why...but this is setting config to be a string (e.g. "{}") whereas the previous code is correctly saving it as JSON in the database. 

P.S. Possibly after we fix this it is worth adding a check that config is JSON to the test?

dadmobile (changes requested):

transformerlab/db.py
aliasaria: Fixed thank you for finding this!

dadmobile (approved):

PR #334, opened 2025-06-02
show model logs in console

mina-parham:


transformerlab/shared/shared.py
```diff
@@ -209,6 +209,7 @@ async def async_run_python_daemon_and_update_status(
         ):
             venv_path = os.path.join(plugin_location, "venv")
             print(f">Plugin has virtual environment, activating venv from {venv_path}")
+            # log.write(f">Plugin has virtual environment, activating venv from {venv_path}")
```
dadmobile, line 212: Uncomment or remove?

transformerlab/shared/shared.py
```diff
@@ -289,7 +291,7 @@ async def run_job(job_id: str, job_config, experiment_name: str = "default", job
         # update task to be marked as COMPLETE:
         await db.job_update_status(job_id, "COMPLETE")
         # implement rest later
-        return {"status": "complete", "job_id": job_id, "message": "Task job completed successfully"}
+        return
```
dadmobile, line 294: I don't understand why you are removing the return value?

transformerlab/shared/shared.py
```diff
@@ -304,19 +306,8 @@ async def run_job(job_id: str, job_config, experiment_name: str = "default", job
...
-            return {"status": "error", "job_id": job_id, "message": "Evaluation job failed: No job data found"}
-        
-        if job_data.get("stop", False):
-            await db.job_update_status(job_id, "STOPPED")
-            return {"status": "stopped", "job_id": job_id, "message": "Evaluation job was stopped by user"}
-        else:
-            await db.job_update_status(job_id, "COMPLETE")
-            return {"status": "complete", "job_id": job_id, "message": "Evaluation job completed successfully"}
+        await db.job_update_status(job_id, "COMPLETE")
+        return
```
dadmobile, line 310: You can't remove this from the database.  We need it for hte app to know if the job succeeded.

dadmobile (changes requested):
I think we were not on the same page.  Will chat with you about what we are looking to do.

PR #335, opened 2025-06-02
Recipes create Eval and Generate Tasks

aahaanmaini:


transformerlab/routers/recipes.py
```diff
@@ -226,6 +226,9 @@ async def create_experiment_for_recipe(id: int, experiment_name: str):
             }
             break  # Only set the first model dependency
 
+    # Ensure experiment always has an adaptor field to prevent KeyError in generation/eval scripts
+    await experiment_router.experiments_update_config(experiment_id, "adaptor", "")
```
aliasaria, line 230: What is the thing about always having an adaptor script?

aliasaria (changes requested):

aliasaria (approved):

PR #336, opened 2025-06-02
Added search and install routes for model adaptors

Sourenm:


deep1401 (approved):
To be merged with the frontend only, don't merge before that

PR #341, opened 2025-06-04
Internal Export Experiment endpoint

aahaanmaini:


transformerlab/routers/experiment/experiment.py
```diff
@@ -183,3 +184,85 @@ async def experiment_get_file_contents(id: str | int, filename: str):
         return ""
 
     return file_contents
+
+
+@router.get("/{id}/export", summary="Export experiment to JSON format", tags=["experiment"])
```
aliasaria, line 189: we have to name this to something like "export_to_recipe" because this is not a true export function (it doesn't export everything in the experiment for example documents or job runs and their outputs, just the parts that are necessary for a recipe)

aliasaria (changes requested):

aliasaria (approved):

PR #342, opened 2025-06-05
added the support for 'valid' splits

Sourenm:


deep1401 (approved):
Could you add somewhere on frontend about these instructions?

PR #347, opened 2025-06-06
Fix tag used to get docker image in dockerfile

dadmobile:
I'm not sure why but our docker image is now tagged with a v before the version so our dockerfile doesn't work anymore!

### DIFF
--- a/docker/cpu/docker-compose.yml.tpl
+++ b/docker/cpu/docker-compose.yml.tpl
@@ -1,6 +1,6 @@
 services:
   transformerlab-api:
-    image: transformerlab/api:${VERSION}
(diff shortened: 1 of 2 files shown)


deep1401 (approved):

deep1401:
http://github.com/transformerlab/transformerlab-api/blob/main/docker/gpu/nvidia/deploy.sh
This file does remove the 'v' , maybe we make the change here?

dadmobile:
OK yeah so instead of adapting to the v being there it looks like we should figure out why the v is there now. Debugging github workflows is annoying!

dadmobile:
Deep found the root cause.

PR #357, opened 2025-06-12
support for schedulers in diffusion

Sourenm:


deep1401 (changes requested):
Can you save the scheduler name to history as well and we need to show that in Model Info on frontend

transformerlab/routers/diffusion.py
```diff
@@ -594,6 +611,22 @@ def get_pipeline(
...
+
+        # This will trap missing keys
+        try:
+            if scheduler != "default":
+                scheduler_class = scheduler_map[scheduler]
+                pipe.scheduler = scheduler_class.from_config(pipe.scheduler.config)
+                log_print(f"[DEBUG] Set scheduler to: {type(pipe.scheduler).__name__}")
+        except KeyError:
+            log_print(f"[ERROR] Unknown scheduler: {scheduler}")
+            raise
```
deep1401, line 624: Could you just instead not set the scheduler instead of raising an error here and just log this? This is what we do in the LoRA part as well

deep1401 (changes requested):
Just one tiny change, the rest is all well done

deep1401 (approved):

PR #360, opened 2025-06-13
Fix Cuda docker and make it so that we only release one dockerfile from common

deep1401:
The install dependencies part is moved to when we actually run the container so we don't need to actually have separate docker containers now


dadmobile (approved):

dadmobile (approved):
I still need to test this a bit but approving because it'll be easier if we get this in the next build and it wasn't working right before anyways.

PR #361, opened 2025-06-13
Move Workflows router under Experiment

aahaanmaini:


transformerlab/db.py
```diff
@@ -990,13 +1001,23 @@ async def workflow_run_get_by_id(workflow_run_id):
...
-    await db.execute(
-        "UPDATE workflows SET status = 'DELETED', updated_at = CURRENT_TIMESTAMP WHERE id = ?", (workflow_id,)
-    )
+    if experiment_id is not None:
+        # Delete with experiment_id check to enforce relationship at DB level
+        result = await db.execute(
+            "UPDATE workflows SET status = 'DELETED', updated_at = CURRENT_TIMESTAMP WHERE id = ? AND experiment_id = ?",
+            (workflow_id, experiment_id),
+        )
+        if result.rowcount == 0:
```
dadmobile, line 1012: Move this logic to the bottom of the function so that the function always behaves the same regardless of if you supply an experiment or not.

dadmobile (changes requested):
Just some small inconsistencies in response to type to fix.

I'll chat with you quickly to make sure we're on the same page about the architecture, but I assume the core principle was to force including experiment at DB level so eventually we'll be able to authorize access to everything by experiment.

dadmobile:
In general this looks fine functionally.  The only thing I would have done differently is I think I would have made most of these DB functions require experiment ID?  Having it optional feels like leaving the door open for mistakes down the road.

dadmobile (changes requested):

transformerlab/db.py
```diff
@@ -990,13 +995,17 @@ async def workflow_run_get_by_id(workflow_run_id):
...
+    # Delete with experiment_id check to enforce relationship at DB level
+    result = await db.execute(
+        "UPDATE workflows SET status = 'DELETED', updated_at = CURRENT_TIMESTAMP WHERE id = ? AND experiment_id = ?",
+        (workflow_id, experiment_id),
     )
+    if result.rowcount == 0:
+        return False  # Workflow not found or doesn't belong to experiment
     await db.commit()
-    return
+    return True
```
dadmobile, line 1008: Could simplify with return result.rowcount > 0

transformerlab/db.py
```diff
@@ -964,8 +964,13 @@ async def workflow_run_get_all():
...
 
-async def workflows_get_by_id(workflow_id):
-    cursor = await db.execute("SELECT * FROM workflows WHERE id = ? ORDER BY created_at desc LIMIT 1", (workflow_id,))
+async def workflows_get_by_id(workflow_id, experiment_id):
+    # Query with both workflow_id and experiment_id to enforce relationship at DB level
+    cursor = await db.execute(
+        "SELECT * FROM workflows WHERE id = ? AND experiment_id = ? AND status != 'DELETED' ORDER BY created_at desc LIMIT 1",
+        (workflow_id, experiment_id),
+    )
+
```
dadmobile, line 973: Do me a favour and change the code (not yours) below this to just return row while you're in here fixing.

PR #362, opened 2025-06-13
Diffusion ControlNets

Sourenm:
Added the support for OpenPose and Canny ControlNets with preprocessing of input images to ControlNets

Note: needs controlnet_aux package to run
This should be added to install.sh routine later


transformerlab/shared/download_huggingface_model.py
```diff
@@ -64,6 +66,22 @@
 
     print(f"Downloading adaptor {peft} with job_id {job_id}")
 
+elif mode == "controlnet":
```
deep1401, line 69: There should be no separate mode for controlnet. It should be downloaded exactly like we download a model for foundation tab. It should also be shown there for now once the download is done. After that is done, we need to just have a function in diffusion.py which lists what all controlnet models are available locally. the condition is that the architecture should have the word "controlnet" in them for now. 

After that we should only show those models. No extra directory called "controlnets" is needed and it shouldn't be created. You need to just modify the the condition which looks in model_index.json somewhere to determine whether the model being downloaded is sd model or not and change that to also check config.json and get the architecture from _class_name instead.

deep1401: Start with `lllyasviel/sd-controlnet-canny`, this shows up with the architecture "UnkownDiffusionPipeline", modify where this is assigned such that it instead tries to read from config.json first before assigning that architecture

transformerlab/shared/diffusion_worker.py
```diff
@@ -51,6 +53,24 @@
 }
 
 
+def load_controlnet_model(controlnet_id: str, device: str = "cuda") -> ControlNetModel:
```
deep1401, line 56: All of this should look into present local models, see the comments in download_huggingface_models.py for more details

transformerlab/routers/diffusion.py
```diff
@@ -169,6 +196,80 @@
 UPSCALE_MODEL_LATENT = "stabilityai/sd-x2-latent-upscaler"
 
 
+def preprocess_for_controlnet(input_pil: Image.Image, controlnet_id: str) -> Image.Image:
```
deep1401, line 199: This should not be dependent on name. There are controlnets who have the ability of doing all of them in a single model. This should be based on a drop-down when you upload a reference image.

deep1401 (changes requested):

deep1401:
I still get the mediapipe not installed warning upon starting the api, just want to make sure that we do not need it somewhere?

deep1401 (approved):

PR #367, opened 2025-06-18
Add https flag

MashDevel:
This PR:

- Adds support for an `--https` flag to serve the API over HTTPS using a self-signed certificate.

- Implements logic in `ssl_utils.py` to generate and persist a self-signed cert and key under the workspace directory.

- Updates run.sh to support a `-s` flag for enabling HTTPS from the CLI. Handles both normal and reload modes accordingly.

- Adds tests for certificate generation, reuse, content (subject, SANs), and concurrency safety.

- Moves an import in `dirs.py` to avoid circular import problems.

This is one of multiple changes aimed at making the server safe to expose directly to the internet. Some cloud providers don't support VPNs or make them hard to set up. So this will make it much easier to use Transformer Lab on multiple cloud providers.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -14,6 +14,7 @@ TLABHOST="0.0.0.0"
 PORT="8338"
 
 RELOAD=false
+HTTPS=false
 
 # echo "Your shell is $SHELL"
 # echo "Conda's binary is at ${CONDA_BIN}"
@@ -31,11 +32,12 @@ else
     echo "✅ Conda is installed."
 fi
 
-while getopts crp:h: flag
+while getopts crsp:h: flag
 do
     case "${flag}" in
         c) CUSTOM_ENV=true;;
         r) RELOAD=true;;
+        s) HTTPS=true;;
         p) PORT=${OPTARG};;
         h) TLABHOST=${OPTARG};;
     esac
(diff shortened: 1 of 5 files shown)


run.sh
```diff
@@ -82,7 +84,15 @@ fi
 echo "▶️ Starting the API server:"
 if [ "$RELOAD" = true ]; then
     echo "🔁 Reload the server on file changes"
-    uv run -v uvicorn api:app --reload --port ${PORT} --host ${TLABHOST}
+    if [ "$HTTPS" = true ]; then
+        uv run -v python api.py --https --reload --port ${PORT} --host ${TLABHOST}
```
deep1401, line 88: We wound want to stick to using `uvicorn` for the server instead of running it directly as that is usually recommended while doing containerization as well. 
Would it be possible to preserve the earlier uvicorn logic whenever run.sh is not run with the `-s` flag and make that default while it can still follow the change here you made to run with python api.py to access your custom arg?

MashDevel: Hi Deep!

Yes for sure, makes me think an env var might be a cleaner choice over a custom flag. So we consistently launch using uvicorn, what do you think?

MashDevel: Though, taking a look at the docs it doesn’t mention any issues/limitations launching programatically vs in directly from the CLI (as long as we use the import string style) https://www.uvicorn.org/deployment/#running-programmatically  

This stack overflow thread also suggests no difference between the two (although it’s a few years old) https://stackoverflow.com/questions/63177681/is-there-a-difference-between-running-fastapi-from-uvicorn-command-in-dockerfile

 I’m curious which potential issues with containerization are of concern?

MashDevel: Went with your original suggestion

deep1401: Thanks, I will try it out and add my review to this one

deep1401 (commented):
Tried this and things work with a site not secure warning in the browser which would be expected for this right now? 
Also you might want to add back the `--no-access-log` option when reload is off and we use the uvicorn command for launching so that we don't show each and every access request on the API as these are also copied over to the global log.

MashDevel:
Yes that's correct, with a self signed cert there's no CA to say it's legit, so will give a warning unless we tell the browser to explicitly trust the cert.

Good catch on the `--no-access-log`!, added it back

transformerlab/shared/dirs.py
```diff
@@ -2,11 +2,9 @@
 
 import os
 from pathlib import Path
```
dadmobile, line 4: Tiny favour...I think I screwed up a merge conflict resolution and accidentally left this line that imports Path in when it should have been removed.  It's causing ruff to fail.  Any chance you can just delete this for me?  

Otherwise I think this is good to merge.

dadmobile (changes requested):

PR #369, opened 2025-06-19
Cancel Workflow API endpoint

aahaanmaini:
Response to https://github.com/transformerlab/transformerlab-app/issues/551

### DIFF
--- a/transformerlab/routers/experiment/workflows.py
+++ b/transformerlab/routers/experiment/workflows.py
@@ -358,6 +358,46 @@ async def workflow_runs_get_by_id(workflow_run_id: str, experimentId: int):
     return returnInfo
 
 
+@router.get("/{workflow_run_id}/cancel", summary="Cancel a running workflow")
+async def cancel_workflow_run(workflow_run_id: str, experimentId: int):
+    """
+    Cancel a running workflow by stopping all its current jobs.
+    The workflow execution engine will automatically detect the stopped jobs
+    and transition the workflow to CANCELLED status.
+    """
+    # Get the workflow run
+    workflow_run = await db.workflow_run_get_by_id(workflow_run_id)
+    if not workflow_run:
+        return {"error": "Workflow run not found"}
+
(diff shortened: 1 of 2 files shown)


transformerlab/routers/experiment/workflows.py
```diff
@@ -358,6 +358,46 @@ async def workflow_runs_get_by_id(workflow_run_id: str, experimentId: int):
...
+    current_status = workflow_run.get("status")
+    if current_status not in ["RUNNING", "QUEUED"]:
+        return {"error": f"Cannot cancel workflow in {current_status} status"}
+
+    # Cancel all current jobs - this will trigger automatic workflow cancellation
+    current_job_ids = json.loads(workflow_run.get("current_job_ids", "[]"))
+    cancelled_jobs = []
+
+    for job_id in current_job_ids:
+        await db.job_stop(job_id)  # This sets stop=True in job_data
```
dadmobile, line 388: Does it matter if the job has run already?  It'll just set stop and nothing else will happen?  Probably want to know if some jobs ran.

transformerlab/routers/experiment/workflows.py
```diff
@@ -358,6 +358,46 @@ async def workflow_runs_get_by_id(workflow_run_id: str, experimentId: int):
...
+        cancelled_jobs.append(job_id)
+
+    # The workflow execution engine will automatically detect the stopped jobs
```
dadmobile, line 391: I've tried parsing the start_next_step_in_workflow code and not sure I understand what happens.  Will it keep trying to start jobs and then stop them because Stop=true or will it actually prevent them from starting?
Just wondering if that code should check for there being a stop on the workflow instead of each job or something?

dadmobile (changes requested):

transformerlab/routers/experiment/workflows.py
dadmobile: Confirmed with Aahaan tha completed jobs show correctly.

PR #371, opened 2025-06-20
Add api for chat_template tokenizer

mina-parham:


transformerlab/routers/model.py
```diff
@@ -972,4 +972,28 @@ async def model_import(model: basemodel.BaseModel):
...
+        template = getattr(tokenizer, "chat_template", None)
+        if template:
+            return {
+                "chat_template": template,
+                "source": "tokenizer_config",
+                "error_code": 0
+            }
+    except Exception as e:
+        return {
+            "message": f"{str(e)}",
```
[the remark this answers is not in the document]
dadmobile: Oh aha it did complain about this!  😄

transformerlab/routers/model.py
```diff
@@ -972,4 +972,28 @@ async def model_import(model: basemodel.BaseModel):
 
     print(f"{model.id} imported successfully.")
 
-    return {"status": "success", "data": model.id}
\ No newline at end of file
+    return {"status": "success", "data": model.id}
+
+@router.post("/model/chat_template")
```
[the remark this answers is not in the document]
mina-parham: I used post in case we need to pass more params later, like `trust_remote_code`, but kept it simple for now.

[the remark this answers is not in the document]
dadmobile: You marked as resolved but didn't answer?

transformerlab/routers/model.py
```diff
@@ -972,4 +974,25 @@ async def model_import(model: basemodel.BaseModel):
...
+async def chat_template(model_name: str):
+
+    try:
+        tokenizer = AutoTokenizer.from_pretrained(
```
dadmobile, line 983: Looks like you need to add a test for this endpoint.

transformerlab/routers/model.py
```diff
@@ -972,4 +974,25 @@ async def model_import(model: basemodel.BaseModel):
...
+            model_name,
+            trust_remote_code=True,
+        )
+        template = getattr(tokenizer, "chat_template", None)
```
dadmobile, line 987: What happens with a local model? GGUF?

dadmobile (changes requested):

transformerlab/routers/model.py
mina-parham: yeah that’s going to be a problem with local models (it will show empty in those cases). One solution could be to create a mapping between local model names and their original hf models, and use that to load the corresponding tokenizer. what do you think?

dadmobile: Interesting. That will exist in info.json in the huggingface_repo for models generated in transformerlab. We could use model provenance in teh future.  And I guess for GGUF models we aren't doing training, so it's less important.

PR #375, opened 2025-06-23
Load gradio UIs from plugins in a secondary gradio-only server

aliasaria:


aliasaria:
Closing this for now as there may be better ways to do this.

PR #376, opened 2025-06-24
render chat template with tokenizer

mina-parham:


transformerlab/routers/data.py
```diff
@@ -239,6 +241,7 @@ async def dataset_preview_with_template(
     template: str = "",
     offset: int = Query(0, description="The starting index from where to fetch the data.", ge=0),
     limit: int = Query(10, description="The maximum number of data items to fetch.", ge=1, le=1000),
+    model_name: str = ''
```
dadmobile, line 244: Overloading this function doesn't feel like the right way to do this. At a minimum we need to add documentation explaining what is going on.  I get wanting to use HF's existing functionality for rendering.
Let me look at your frontend PR first to see what we can do.

transformerlab/routers/data.py
```diff
@@ -277,9 +283,15 @@ async def dataset_preview_with_template(
         row["__index__"] = i + offset
         for key in result["columns"].keys():
             row[key] = serialize_row(result["columns"][key][i])
-
-        # Apply the template to a new key in row called __formatted__
-        row["__formatted__"] = jinja_template.render(row)
+        
+        if model_name:
+            row["__formatted__"] = tokenizer.apply_chat_template(
+                    row["messages"],
```
dadmobile, line 289: Wouldn't this only work if there was exactly a column called "messages" in the dataset?

dadmobile (changes requested):

test/api/test_data.py
```diff
@@ -144,3 +148,45 @@ def test_edit_with_template():
...
+    dataset_dict = DatasetDict({
+        "train": Dataset.from_dict({
+            "message": [
+                [{"role": "user", "content": "Hello!"}],
+                [{"role": "user", "content": "What is AI?"}],
+            ]
+        })
+    })
+
+    mock_tokenizer = MagicMock()
```
dadmobile, line 162: I don't know if there's simpler way to do this in light of Ali's comments about trying to keep these tests cleaner and avoid mocks and patches. Is there critical things to test here to make sure we aren't breaking or is this just cover a branch of code?

dadmobile (changes requested):

transformerlab/routers/data.py
```diff
@@ -277,7 +281,7 @@ async def dataset_preview_with_template(
         row["__index__"] = i + offset
         for key in result["columns"].keys():
             row[key] = serialize_row(result["columns"][key][i])
-
+
```
dadmobile, line 284: I bet ruff complains about this tab. :)

dadmobile (changes requested):

dadmobile:
Really sorry this was out for so long but I'm not sure how to resolve the conflicts now because the tests changed. :(

dadmobile (approved):

PR #388, opened 2025-06-29
Fix: HF Cache won't be deleted when user clicks Ctrl+C

ucalyptus:
Closes: https://github.com/transformerlab/transformerlab-app/issues/575


aliasaria (approved):
Thanks!

PR #389, opened 2025-06-30
Delete plugin from workspace on install failure

deep1401:
Fixes transformerlab/transformerlab-app#576

### DIFF
--- a/transformerlab/routers/plugins.py
+++ b/transformerlab/routers/plugins.py
@@ -92,6 +92,18 @@ async def copy_plugin_files_to_workspace(plugin_id: str):
     copy_tree(plugin_path, dirs.plugin_dir_by_name(plugin_id))
 
 
+async def delete_plugin_files_from_workspace(plugin_id: str):
+    plugin_id = secure_filename(plugin_id)
+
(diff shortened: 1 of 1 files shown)


transformerlab/routers/plugins.py
```diff
@@ -92,6 +92,18 @@ async def copy_plugin_files_to_workspace(plugin_id: str):
     copy_tree(plugin_path, dirs.plugin_dir_by_name(plugin_id))
 
 
+async def delete_plugin_files_from_workspace(plugin_id: str):
+    plugin_id = secure_filename(plugin_id)
+
+    plugin_path = os.path.join(dirs.PLUGIN_DIR, plugin_id)
+    # create the directory if it doesn't exist
```
dadmobile, line 99: I think this comment is wrong?

dadmobile (changes requested):
Tiny nitpick. :)

transformerlab/routers/plugins.py
deep1401: Haha yes, sorry I copied over the function from above and forgot to change that. 
Changed now

PR #390, opened 2025-06-30
Make fixture module based to fix db lock in tests

deep1401:


deep1401:
Closing this because Ali's update on main fixed the issue. 
Its either because we skipped all workflow tests or because we increased PRAGMA timeout to 30 seconds from 5 seconds

PR #400, opened 2025-07-02
Added the supported architectures to index.json for fastchat and mlx server plugins

Sourenm:


transformerlab/plugins/mlx_server/index.json
```diff
@@ -34,7 +34,8 @@
     "Phi3ForCausalLM",
     "Qwen2ForCausalLM",
     "Qwen3ForCausalLM",
-    "Qwen3MoeForCausalLM"
+    "Qwen3MoeForCausalLM",
+    "Ernie4_5_ForCausalLM"
```
dadmobile, line 38: MLX doesn't support the MOE model?

Sourenm: I read somewhere that although it does support the MoE variants in theory, they are so huge that it makes them impractical, that's why I did not add them here

dadmobile (approved):

PR #401, opened 2025-07-02
Add/vllm proxy server

mina-parham:


transformerlab/plugins/vllm_proxy_worker/setup.sh
```diff
@@ -0,0 +1,2 @@
+uv pip install vllm==0.5.4
```
dadmobile, line 1: Let's experiment to see if we can get a newer version running!

dadmobile (changes requested):

transformerlab/plugins/vllm_proxy_worker/index.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "vLLM Proxy Server",
```
dadmobile, line 2: And then last thing is...let's ask the group what we should call this server. I don't think this name will mean anything to a user!

dadmobile (changes requested):

transformerlab/fastchat_openai_api.py
```diff
@@ -393,6 +393,7 @@ async def get_gen_params(
         "echo": echo,
         "stream": stream,
         "logprobs": logprobs,
+        "messages": messages,
```
dadmobile, line 396: I'm having trouble parsing this.  Why are you adding messages here?

transformerlab/plugins/vllm_server/main.py
```diff
@@ -23,78 +24,76 @@
...
-#             "gptq_marlin",
-#             "awq_marlin",
-#             "gptq",
-#             "squeezellm",
-#             "compressed-tensors",
-#             "bitsandbytes",
-#             "None"
-#     ]
-# }
+print("Starting OpenAI API Proxy Server", file=sys.stderr)
```
dadmobile, line 27: Hmmm...I know this is maybe technically true but probably you don't need to include this?  I would just not mention "OpenAI" in the output as it might be confusing?

transformerlab/plugins/vllm_server/main.py
```diff
@@ -23,78 +24,76 @@
...
+    "-m",
+    "vllm.entrypoints.openai.api_server",
+    "--model", model,
+    "--dtype", "float16",
+    "--port", str(port),
+    "--gpu-memory-utilization", "0.9",
+    "--enforce-eager",
+    "--trust-remote-code",
+]
+print("Starting vLLM OpenAI API server...", file=sys.stderr)
```
dadmobile, line 56: Similarly I would just change this to remove "OpenAI API" and maybe move it right up to where the other comment is (I realize behind the scenes there are two servers so will leave up to you).

dadmobile (changes requested):

dadmobile (approved):

PR #402, opened 2025-07-02
Fix/sqlalchemy fixes

deep1401:


transformerlab/db/utils.py
```diff
@@ -0,0 +1,50 @@
...
+        obj: SQLAlchemy model instance
+        
+    Returns:
+        dict: Dictionary representation of the object without SQLAlchemy internals
+    """
+    if obj is None:
+        return None
+        
+    # Fields to exclude from the dictionary
+    excluded_fields = {'created_at', 'updated_at'}
```
dadmobile, line 25: Why are you excluding these?

deep1401: That is because these aren't necessary to include and we set start_time and end_time based on the job data. Having these now means having to deal with datetime data because of SQLAlchemy's way of storing them

dadmobile (approved):

PR #407, opened 2025-07-03
Fix Huggingface TRL Trainer on the AMD Machine

deep1401:
- Fixes HIP memory issues on AMD GPUs

### DIFF
--- a/transformerlab/plugins/llama_trainer/main.py
+++ b/transformerlab/plugins/llama_trainer/main.py
@@ -15,8 +15,6 @@ if shutil.which("rocminfo") is not None:
     os.environ["TOKENIZERS_PARALLELISM"] = "false"
 if torch.cuda.is_available():
     os.environ["CUDA_VISIBLE_DEVICES"] = "0"
-    if not HAS_AMD:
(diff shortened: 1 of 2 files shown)


transformerlab/plugins/llama_trainer/main.py
```diff
@@ -8,9 +8,15 @@
...
     HAS_AMD = True
+    # AMD-specific optimizations
+    os.environ["PYTORCH_HIP_ALLOC_CONF"] = "max_split_size_mb:128"
+    os.environ["HIP_VISIBLE_DEVICES"] = "0"
+    # Disable some problematic CUDA-specific features
+    os.environ["TOKENIZERS_PARALLELISM"] = "false"
 if torch.cuda.is_available():
     os.environ["CUDA_VISIBLE_DEVICES"] = "0"
-    os.environ["HIP_VISIBLE_DEVICES"] = "0"
+    if not HAS_AMD:
```
dadmobile, line 18: Just double checking that this not is supposed to be here? Possibly because you set it in the rocminfo above for AMD cases? Maybe a comment to explain would be helpful.

PR #410, opened 2025-07-04
Add Docker for AMD GPUs

deep1401:
- We do not have a ROCM image for transformerlab-api
- This is the Dockerfile anyone can use to create the ROCM docker image for transformer lab
- The `entrypoint.sh` file was added because doing a direct execution with bash -c for entrypoint kept throwing errors even if the file system existed correctly
- Modified Github action to also release a `version-rocm` tagged image for rocm specifically


deep1401:
> Looks good but I'm not testing this because I don't know how. We'll do a build next week and fix it if it doesn't work. :P

This is tested locally. We'll test the action upon release

PR #412, opened 2025-07-05
Changes required for RAG pipeline recipe

aahaanmaini:
These are all the changes required for the RAG pipeline recipe, which includes training a custom embedding model and then evaluating both. Majority of these changes are required for the DeepEval LLM-as-Judge plugin.

Summary of changes and why they are important:
- Improved JSON Repair Pipeline: Added 3-step fallback system to fix malformed JSON from small models, including cleanup of tokens, markdown, and quotes.
- Pydantic Schema Fallback: Built robust schema handling compatible with Pydantic v1/v2 to prevent DeepEval crashes during parsing.
- Schema Compliance Criticality: DeepEval relies on exact schemas for metrics; without compliance, evaluations break deep in the pipeline.
- Prompt Enhancements: Added logic to detect JSON tasks and auto-insert formatting instructions to improve model output.
- Dependency Note: Updated dependencies for langchain

### DIFF
--- a/transformerlab/plugin_sdk/transformerlab/sdk/v1/tlab_plugin.py
+++ b/transformerlab/plugin_sdk/transformerlab/sdk/v1/tlab_plugin.py
@@ -7,7 +7,7 @@ import traceback
 import requests
 import json
 from pydantic import BaseModel
-from typing import Any, List
+from typing import Any, List, Type
 
 from datasets import get_dataset_split_names, get_dataset_config_names, load_dataset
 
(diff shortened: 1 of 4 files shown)


deep1401 (changes requested):
The right way to do this is to have json mode within Fastchat and not do it like this. 
Not sure if there were any discussions where Tony asked you to do this one specifically but we don't want to do so much custom parsing as we also use these functions in a lot of different plugins

aliasaria:
Closing this for now as we need to find a different less complicated way (probably supporting structured output in inference better)

PR #414, opened 2025-07-07
Name outputs of Generation tasks

aahaanmaini:


deep1401:
Not sure what the end goal is here but I'm pretty sure the name "dataset_name" is reserved and should not be added here as that is what we use within tasks

aliasaria (approved):

aliasaria (changes requested):
Let's chat with Deep to address his concerns first

transformerlab/plugin_sdk/transformerlab/sdk/v1/generate.py
```diff
@@ -55,27 +58,26 @@ def save_generated_dataset(self, df, additional_metadata=None, dataset_id=None,
         """
         self._ensure_args_parsed()
 
+        # Use dataset_id priority: explicit parameter > user input > default naming
+        if dataset_id is None:
+            dataset_id = self.params.get("output_dataset_name")
```
deep1401, line 63: Lets not do this here at all, this should be done inside the plugin so the name of the param could be anything and not just output_dataset_name

transformerlab/plugin_sdk/transformerlab/sdk/v1/generate.py
```diff
@@ -55,27 +58,26 @@ def save_generated_dataset(self, df, additional_metadata=None, dataset_id=None,
...
-            if dataset_id is None:
-                dataset_id = f"{self.params.run_name}_{self.params.job_id}".lower()
         else:
             lines = False
             if suffix is not None:
                 output_file = os.path.join(output_dir, f"{self.params.run_name}_{self.params.job_id}_{suffix}.json")
             else:
                 output_file = os.path.join(output_dir, f"{self.params.run_name}_{self.params.job_id}.json")
 
-            if dataset_id is None:
```
deep1401, line 74: Not sure why this is being removed? We need this to be here for image datasets

deep1401 (changes requested):
All of this should just be done within the plugin and when the logic inside the plugin calls the save_generated_dataset function, you just provide the value from output_dataset_name, we should not be modifying the sdk or changing the logic in there to accomodate a hardcoded field name

deep1401 (approved):

aliasaria (approved):

PR #416, opened 2025-07-07
Change logic for performing json.loads to fix error in workflows

deep1401:


transformerlab/routers/experiment/workflows.py
```diff
@@ -381,7 +380,9 @@ async def workflow_runs_get_by_id(workflow_run_id: str, experimentId: int):
     returnInfo = {"run": workflow_run, "workflow": workflow, "jobs": []}
 
     try:
-        if not isinstance(workflow_run.get("job_ids", "[]"), dict) or not isinstance(workflow_run.get("job_ids", "[]"), list):
+        if not isinstance(workflow_run.get("job_ids", "[]"), dict) or not isinstance(
```
dadmobile, line 383: Possibly also need to update this logic?

dadmobile (changes requested):

transformerlab/routers/experiment/workflows.py
deep1401: Fixed

dadmobile (approved):

PR #421, opened 2025-07-08
Diffusion Plugin

Sourenm:


api.py
```diff
@@ -46,11 +46,10 @@
     prompts,
     tools,
     batched_prompts,
-    diffusion,
     recipes,
     users,
 )
-
+from transformerlab.routers.experiment import diffusion
```
deep1401, line 52: We do not import it like this, look at the routers/experiment/experiment.py file inside and see how other routes under experiment are imported and follow that structure

transformerlab/plugin_sdk/transformerlab/plugin.py
```diff
@@ -452,6 +452,10 @@ def set_job_completion_status(
             self.add_to_job_data("completion_status", completion_status)
             self.add_to_job_data("completion_details", completion_details)
             
+            # Update the job status field if there's a failure
+            if completion_status == "failed":
```
deep1401, line 456: Is this a duplicate change? 
Maybe remove this as we do the same on the next line

transformerlab/plugin_sdk/transformerlab/sdk/v1/diffusion.py
```diff
@@ -0,0 +1,82 @@
...
+        key = None
+        for arg in unknown_args:
+            if arg.startswith("--"):
+                key = arg.lstrip("-")
+                self.params[key] = True
+            elif key:
+                self.params[key] = arg
+                key = None
+
+    def save_generated_images(self, metadata: dict = None, suffix: str = None):
```
deep1401, line 38: Not sure what this is being used for? If its a dataset then it is valid but if you're trying to save a generated image with this then this might be wrong?

transformerlab/plugin_sdk/transformerlab/sdk/v1/diffusion.py
```diff
@@ -0,0 +1,82 @@
...
+        with open(metadata_file, "w") as f:
+            json.dump(metadata or {}, f, indent=2)
+
+        print(f"[DIFFUSION] Metadata saved to: {metadata_file}")
+        return metadata_file
+
+    def get_output_file_path(self, suffix="", dir_only=False):
+        self._ensure_args_parsed()
+
+        workspace_dir = os.environ.get("_TFL_WORKSPACE_DIR", "./")
```
deep1401, line 65: Instead of this could you import WORKSPACE_DIR from transformerlab.plugin and use that?

transformerlab/plugin_sdk/transformerlab/sdk/v1/diffusion.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            metadata_file = os.path.join(output_dir, f"{self.params.run_name}_{self.params.job_id}_metadata.json")
+
+        with open(metadata_file, "w") as f:
+            json.dump(metadata or {}, f, indent=2)
+
+        print(f"[DIFFUSION] Metadata saved to: {metadata_file}")
+        return metadata_file
+
+    def get_output_file_path(self, suffix="", dir_only=False):
```
deep1401, line 62: I understand why we add this here but if this is the exact same with only a difference of "diffusion" and "datasets", then it would be wiser to move this out to the base TLabPlugin class and add a parameter so GenTLab and DiffusionTLab can both use

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -430,24 +262,62 @@ class ImageHistoryItem(BaseModel):
...
+    controlnet_model = ControlNetModel.from_pretrained(
+        controlnet_id, torch_dtype=torch.float16 if device != "cpu" else torch.float32
+    )
+    return controlnet_model
 
 
-# Global cache for loaded pipelines
-# _PIPELINES: dict = {}
-_PIPELINES_LOCK = threading.Lock()
+def _setup_diffusion_logger():
```
deep1401, line 275: Since this is a plugin now, can't we just print instead of doing logger? The output would now show up under output.txt anyway

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -430,24 +262,62 @@ class ImageHistoryItem(BaseModel):
...
 
-# History file path
-HISTORY_FILE = "history.json"
+    if not WORKSPACE_DIR:
+        raise RuntimeError("Missing _TFL_WORKSPACE_DIR")
+
+    # Get the absolute parent directory of WORKSPACE_DIR
+    HOME_DIR = os.path.abspath(os.path.join(WORKSPACE_DIR, ".."))
+    GLOBAL_LOG_PATH = os.path.join(HOME_DIR, "transformerlab.log")
+    logger = logging.getLogger("diffusion")
```
deep1401, line 285: We should not have this here, just do a print within the plugin as the logic needs to be as simple as possible

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -720,7 +483,11 @@ def get_pipeline(
         # Load LoRA adaptor if provided - same code for local and HF Hub!
         if adaptor and adaptor.strip():
             try:
-                adaptor_dir = os.path.join(os.environ.get("_TFL_WORKSPACE_DIR"), "adaptors", secure_filename(model))
+                adaptor_dir = os.path.join(
+                    os.environ.get("_TFL_WORKSPACE_DIR"),
```
deep1401, line 487: This should be imported from transformerlab.plugin as an earlier comment

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -1097,7 +1160,22 @@ def run_pipe():
             generation_start = time.time()
             log_print("Starting image generation...")
 
+            progress_queue = queue.Queue()
+
+            async def monitor_progress():
```
deep1401, line 1165: What is this used for? I thought we do progress update inside the callback

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -1214,889 +1293,85 @@ def run_upscale():
...
-            if "flux" in arch.lower():
-                return True
+        log_print(f"Error during image generation: {str(e)}")
+        raise HTTPException(status_code=500, detail=f"Image generation failed: {str(e)}")
 
-        return False
-    except Exception:
-        return False
 
+def parse_args():
```
deep1401, line 1320: We should not be parsing any arg inside a plugin

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -1214,889 +1293,85 @@ def run_upscale():
...
-    """
-    Lists all downloaded ControlNet models by reading the controlnets directory
-    and extracting `_class_name` from their config.json.
-    """
-    all_models = await model_helper.list_installed_models(embedding=False)
-    models = []
+    parser.add_argument("--plugin_dir", type=str, required=True)
+    parser.add_argument("--job_id", type=str, required=True)
+    parser.add_argument("--experiment_name", type=str, required=True)
+    parser.add_argument("--run_name", type=str, default="diffused")
```
deep1401, line 1326: I guess this is stale code and you need to remove it?

transformerlab/shared/shared.py
```diff
@@ -410,7 +411,115 @@ async def run_job(job_id: str, job_config, experiment_name: str = "default", job
...
+                f.write("")
+
+        plugin_dir = dirs.plugin_dir_by_name(plugin_name)
+        plugin_main_args = ["--plugin_dir", plugin_dir]
+
+        # Flatten job_config["config"] into CLI args
+        config = job_config.get("config", {})
+
+        # Convert base64 images to files and update config
+        base64_fields = {
```
deep1401, line 438: Why do you we need to add this to config? Why cant the plugin directly fetch them?

deep1401 (changes requested):

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -1214,889 +1293,85 @@ def run_upscale():
...
-        is_controlnet = json_data.get("is_controlnet", False)
+    return args, config
 
-        if (arch and "controlnet" in arch.lower()) or is_controlnet:
-            models.append(model)
 
-    return {"controlnets": models}
+def main():
+    print("Starting image diffusion...", flush=True)
+    asyncio.run(async_main())
```
deep1401, line 1347: Why do we need to do this? And if you do need this, try to see how we have a async_job_wrapper

transformerlab/plugin_sdk/transformerlab/plugin.py
Sourenm: the duplicate is removed.

transformerlab/plugin_sdk/transformerlab/sdk/v1/diffusion.py
Sourenm: solved (comment above)

transformerlab/plugins/image_diffusion/main.py
Sourenm: Done

transformerlab/plugins/image_diffusion/main.py
Sourenm: done

transformerlab/plugins/image_diffusion/main.py
Sourenm: removed

transformerlab/plugins/image_diffusion/main.py
Sourenm: modified

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -0,0 +1,1261 @@
...
+            os.remove(config_path)
+        except Exception:
+            pass
+        raise e
+
+
+@tlab_diffusion.async_job_wrapper(progress_start=0, progress_end=100)
+async def diffusion_generate_job():
+    job_config = {
+        "plugin": tlab_diffusion.params.get("plugin", "image_diffusion"),
```
deep1401, line 785: We probably don't need this?

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -0,0 +1,1261 @@
...
+        try:
+            os.remove(config_path)
+        except Exception:
+            pass
+        raise e
+
+
+@tlab_diffusion.async_job_wrapper(progress_start=0, progress_end=100)
+async def diffusion_generate_job():
+    job_config = {
```
deep1401, line 784: Why aren't we directly using the tlab_diffusion.params instead of maintaining this?

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -0,0 +1,1261 @@
...
+        "is_controlnet": tlab_diffusion.params.get("is_controlnet", ""),
+        "save_intermediate_images": tlab_diffusion.params.get("save_intermediate_images", True),
+    }
+    # Convert image files back to base64 if present
+    image_path_keys = {
+        "input_image_path": "input_image",
+        "mask_image_path": "mask_image",
+    }
+
+    for path_key, base64_key in image_path_keys.items():
```
deep1401, line 821: Just to clarify here on previous comment, this makes the images back to base64 for processing by diffusers right?

transformerlab/plugins/image_diffusion/setup.sh
deep1401, line 1: Just a note: we left this blank for now because all files are being installed by the original requirements

transformerlab/plugins/image_diffusion/main.py
```diff
@@ -0,0 +1,1261 @@
+from fastapi import HTTPException
+from pydantic import BaseModel
+from huggingface_hub import model_info
+import base64
+from io import BytesIO
+import torch
+import asyncio
+import threading
+import gc
+from transformerlab.plugin import WORKSPACE_DIR
```
deep1401, line 10: Maybe we standardize using this throughtout the file?

transformerlab/plugin_sdk/transformerlab/sdk/v1/diffusion.py
```diff
@@ -0,0 +1,39 @@
...
+    """Enhanced decorator class for TransformerLab diffusion plugins"""
+
+    def __init__(self):
+        super().__init__()
+        self._parser.add_argument("--run_name", default="diffused", type=str, help="Name for the diffusion output")
+        self._parser.add_argument("--experiment_name", default="default", type=str, help="Name of the experiment")
+        self._parser.add_argument("--diffusion_model", default="local", type=str, help="Diffusion model to use")
+        self._parser.add_argument("--model", type=str, default="")
+        self._parser.add_argument(
+            "--diffusion_type", default="txt2img", type=str, help="Type of diffusion task (txt2img, img2img, etc.)"
```
deep1401, line 14: Do we use this param somewhere?

transformerlab/shared/shared.py
deep1401, line 1: Adding it here because its probably not included in the diff, now that we include diffusion_worker in the plugin, we should not have a copy in the shared folder

transformerlab/shared/shared.py
```diff
@@ -410,7 +411,105 @@ async def run_job(job_id: str, job_config, experiment_name: str = "default", job
...
+                    "message": "Diffusion job completed successfully",
+                }
+            else:
+                await db_jobs.job_update_status(job_id, "FAILED")
+                print(f"[DIFFUSION] Job {job_id} failed with return code {process.returncode}")
+                return {"status": "error", "job_id": job_id, "message": "Diffusion job failed"}
+        except Exception as e:
+            await db_jobs.job_update_status(job_id, "FAILED")
+            print(f"[DIFFUSION] Job {job_id} execution error: {e}")
+            return {"status": "error", "job_id": job_id, "message": "Diffusion job failed"}
```
deep1401, line 510: For error returning, there are some cases like when using a SDXL model and the height and width are not multiples of 8, this will error out but we want to show a correct error on why it errored out. The current implementation prints out the error but maybe we could send out the HTTPResponse for the error somehow and show that via notifications? 

Additionally, we will also need to stream output from this job, as it is shown for other jobs, maybe we include a button where we show the progress to see the raw output of the generation which uses the same ViewOutputStreamingModal on the app side?

deep1401: Either we do the HTTPResponse thing or we make use of how the errors are reported on /generate right now in the except block and we show the errors from the job within there as well if a job has failed status

transformerlab/routers/experiment/diffusion.py
```diff
@@ -0,0 +1,1126 @@
...
+from transformerlab.shared import dirs
+from transformerlab.shared.shared import slugify
+import transformerlab.db.jobs as db_jobs
+import logging
+import sys
+
+
+router = APIRouter(prefix="/diffusion", tags=["diffusion"])
+
+UNIVERSIAL_GENERATION_ID = str(uuid.uuid4())
```
deep1401, line 22: I don't know how we reached to using this but this should not be used here. 
This is why we have the /generate_id route which generates a generation_id beforehand and you either use that or one provided with the /generate but the `UNIVERSIAL_GENERATION_ID ` should not be used

deep1401 (changes requested):

transformerlab/plugins/image_diffusion/main.py
Sourenm: Done

transformerlab/plugins/image_diffusion/main.py
Sourenm: Yes that's correct

transformerlab/plugins/image_diffusion/main.py
Sourenm: Done

transformerlab/plugin_sdk/transformerlab/sdk/v1/diffusion.py
Sourenm: Removed

deep1401 (changes requested):
This is almost ready to merge, just a couple of small changes to ensure we still support the plugins we created surrounding this still work

dadmobile:
This worked for me!

dadmobile (approved):

deep1401:
> Was there any context on the decision to put this in a separate kind of plugin section? It took me a while to figure out there was a separate diffusion plugin section. :)

The only context was that it would be nicer when we have other diffusion specific plugins like video or mlx. I could just shift it into loader plugins?

dadmobile (approved):

deep1401 (approved):

PR #424, opened 2025-07-09
Deleting experiment also deletes associated tasks and other things

deep1401:


deep1401:
> I think that this would be better if the marking as DELETED wasn't done at the db layer, but one layer above. So instead of directly modifying the related entities in SQL, you would just call their respective db methods for delete, and if we change the way we handle DELETE in the future, this will still work.

Okay some of the functions to get by experiment id didn't exist in the db and we didn't get have a workflow run delete method too. Added those and switched the logic now to use the original delete functions so the logic is dictated by those instead

PR #430, opened 2025-07-10
Upgrade transformers version

mina-parham:


deep1401 (commented):
I think we used to do this earlier and the command to add things was a bit different in uv. 
It used to show up as something like this in the requirements file upon doing uv pip compile:

```
fschat @ git+https://github.com/transformerlab/FastChat
    # via -r requirements.in
   ```

Is the new thing added to the requirements-uv a changed way of how uv does it?

File: requirements-no-gpu-uv.txt [hunk 1]
```diff
@@ -536,7 +536,7 @@ tqdm==4.66.5
     #   transformers
 transformerlab-inference==0.2.39
     # via -r requirements.in
-transformers==4.53.1
+transformers @ git+https://github.com/huggingface/transformers.git@571a8c21313ef734e77cf9874ea0334d25bd7ff5
```
deep1401, line 539: A commit hash is usually added by uv but we used to remove this so it doesn't have issues with uv pip sync

requirements-uv.txt
```diff
@@ -574,13 +536,11 @@ tqdm==4.66.5
     #   transformers
 transformerlab-inference==0.2.39
     # via -r requirements.in
-transformers==4.53.1
+transformers @ git+https://github.com/huggingface/transformers.git@571a8c21313ef734e77cf9874ea0334d25bd7ff5
```
deep1401, line 539: Same here too

deep1401 (approved):
Tested this on fastchat with llama 3.2 1B and smollm3 3B (enabling unsupported mode) and it works. 
Tested on llama_trainer and llama_trainer_multi_gpu and nothing breaks

PR #435, opened 2025-07-10
Fix Generation jobs showing Complete when Failed

aahaanmaini:


transformerlab/shared/shared.py
```diff
@@ -353,8 +353,14 @@ async def run_job(job_id: str, job_config, experiment_name: str = "default", job
             await db_jobs.job_update_status(job_id, "STOPPED")
             return {"status": "stopped", "job_id": job_id, "message": "Generation job was stopped by user"}
         else:
-            await db_jobs.job_update_status(job_id, "COMPLETE")
-            return {"status": "complete", "job_id": job_id, "message": "Generation job completed successfully"}
+            # Only mark as COMPLETE if progress is 100%
+            progress = job_row.get("progress", -1)
+            if progress == 100:
+                await db_jobs.job_update_status(job_id, "COMPLETE")
+                return {"status": "complete", "job_id": job_id, "message": "Generation job completed successfully"}
```
aliasaria, line 360: @deep1401 is this a good way to check if the generation happened? I don't know the details of of how plugins + progress work but is there a chance that progress is not updated in the plugin, but the generation still completes? What is the best way to check if it was successful?

aliasaria (changes requested):

transformerlab/shared/shared.py
deep1401: I think a better way of doing this is to report errors within the plugin, I think there might be some plugins where errors aren't being raised correctly. If you raise an error inside a plugin then the sdk marks it as FAILED on its own

aliasaria (approved):

PR #450, opened 2025-07-14
centralize worker PID file writing

MashDevel:
for issue https://github.com/transformerlab/transformerlab-app/issues/619

### DIFF
--- a/test/plugin_sdk/test_plugin.py
+++ b/test/plugin_sdk/test_plugin.py
@@ -1,0 +1,26 @@
+import os
+import tempfile
+import pytest
+from unittest import mock
+
+from transformerlab.plugin_sdk.transformerlab import plugin
+
+def test_register_process_single_and_multiple_pids():
+    with tempfile.TemporaryDirectory() as temp_dir:
+        pid_file = os.path.join(temp_dir, "worker.pid")
+        with mock.patch.dict(os.environ, {"LLM_LAB_ROOT_PATH": temp_dir}):
+            pids = plugin.register_process(12345)
+            assert pids == [12345]
+            with open(pid_file) as f:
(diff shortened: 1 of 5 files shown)


test/plugin_sdk/test_plugin.py
```diff
@@ -0,0 +1,26 @@
+import os
+import tempfile
+import pytest
+from unittest import mock
+
+from transformerlab.plugin_sdk.transformerlab import plugin
+
+def test_register_process_single_and_multiple_pids():
```
deep1401, line 8: We're trying to avoid using mock-patch ideology as much as we can while writing tests as we recently undid a huge block of tests which used those. Maybe we could test this in a simpler way?

transformerlab/plugin_sdk/transformerlab/plugin.py
```diff
@@ -17,6 +17,24 @@
...
+def register_process(pid_or_pids):
+    """
+    Record one or many PIDs in <LLM_LAB_ROOT_PATH>/worker.pid so that
+    the 'worker_stop' endpoint can later clean them up.
+    """
+    if isinstance(pid_or_pids, int):
+        pids = [pid_or_pids]
+    else:
+        pids = list(pid_or_pids)
+    root_dir = os.getenv("LLM_LAB_ROOT_PATH")
```
dadmobile, line 29: I hate that we have to do this in the code because we don't have access to the dirs constant here. Since this happens elsewhere in this file (in an inconsistent way), can you do me a huge favour and open an issue saying "figure out how to better access APi directories in plugin harness" or something?

dadmobile (approved):
Deep will have to approve the test fix.

deep1401 (approved):

PR #451, opened 2025-07-15
Remove Training Templates in API

aliasaria:
Since we are moving to Recipes being a global concept, we no longer need the concept of Training Templates. This change removes them from the API


dadmobile (changes requested):
Trying to verify but I'd feel more comfortable if we didn't have any references to these endpoints in the frontend. It looks mostly clear although I found this in endpoints.json (not sure if it's actually used though?):
```
export function GET_TRAINING_TEMPLATE_URL() {
  return `${API_URL()}train/templates`;
}
```

PR #452, opened 2025-07-15
Improve error message for invalid or private Hugging Face dataset IDs

mina-parham:
Fixes: #617

https://github.com/transformerlab/transformerlab-app/issues/617


deep1401 (changes requested):
Maybe we should check for RepositoryNotFoundError or whichever error it raises instead of checking for explicit text?

mina-parham:
> Maybe we should check for RepositoryNotFoundError or whichever error it raises instead of checking for explicit text?

deep1401 (approved):
Tested with `touch-rugby-evals` vs `transformerlab/touch-rugby-evals`

PR #459, opened 2025-07-15
Add a str() guard for generation params

Sourenm:


deep1401 (approved):
A comment here that we do this purely for WD14 which ends up sending some float values which aren't permitted

PR #465, opened 2025-07-17
Add/vllm image

mina-parham:


transformerlab/plugins/vllm_server/main.py
```diff
@@ -92,4 +117,5 @@
     print(line, file=sys.stderr)
 
 print("Vllm worker exited", file=sys.stderr)
+clear_vram()
 sys.exit(1)  # 99 is our code for CUDA OOM
```
dadmobile, line 121: Why are we doing exit code 1.  And we can delete the unrelated comment?

dadmobile (changes requested):

PR #468, opened 2025-07-17
Fix mlx lora trainer bug for str iters

deep1401:


deep1401:
This is a placeholder PR. If we decide to include #440 in the build then this should NOT be merged. 
If not, this needs to go in the build

PR #471, opened 2025-07-18
Isolate Jobs to Experiment

aahaanmaini:


deep1401 (changes requested):
This is not the complete issue as the PR only restricts some parameters to experiment_id.
All tasks/jobs within our app right now are created with the correct experiment_id. The real problem is that when you create a workflow and use it within a recipe the task with that particular experiment_id gets saved in the config and importing it might create issues. The frontend just filters jobs currently without providing any experiment_id. 

To solve this, we might have to solve how tasks work within workflows and how recipe workflows are imported.

aahaanmaini:
Closing since the changes are implemented as part of moving the jobs router under experiment.

PR #483, opened 2025-07-25
Move Jobs Router under Experiment

aahaanmaini:


api.py
```diff
@@ -361,10 +361,12 @@ async def server_worker_start(
     if exitcode is not None and exitcode != 0:
         with open(dirs.GLOBAL_LOG_PATH, "a") as global_log:
             global_log.write(f"Error loading model: {model_name} with exit code {exitcode}\n")
-        error_msg = await job_get_error_msg(job_id)
+        job = await job_create(job_id)
```
dadmobile, line 364: I don't understand why you are creating a new job here?

transformerlab/models/model_helper.py
```diff
@@ -30,11 +30,11 @@ async def is_model_installed(model_id: str):
     return await localmodel.LocalModelStore().has_model(model_id)
 
 
-async def list_model_provenance(model_id: str):
+async def list_model_provenance(model_id: str, experiment_id: int):
```
dadmobile, line 33: I think it's going to be a problem needing experiment_id on this.  It means you can't get model provenance if you aren't in the right experiment?

transformerlab/routers/model.py
```diff
@@ -603,21 +628,30 @@ async def download_gguf_file_from_repo(model: str, filename: str, job_id: int |
     except Exception as e:
         error_msg = f"Error accessing model repository: {type(e).__name__}: {e}"
         if job_id:
-            await db_jobs.job_update_status(job_id, "FAILED", error_msg)
+            job = await db_jobs.job_get(job_id)
```
dadmobile, line 631: I don't think you should be adding this.  We already have experiment_id getting passed to the function.  Won't the DB function just not update if we call it and it doesn't match the job?  Which is the same result?

transformerlab/routers/model.py
```diff
@@ -603,21 +628,30 @@ async def download_gguf_file_from_repo(model: str, filename: str, job_id: int |
...
+            if job is not None:
+                experiment_id = job["experiment_id"]
+                await db_jobs.job_update_status(job_id, "FAILED", error_msg, experiment_id)
         return {"status": "error", "message": error_msg}
 
     if model_details is None:
         error_msg = f"Error reading config for model with ID {model}"
         if job_id:
-            await db_jobs.job_update_status(job_id, "FAILED", error_msg)
+            job = await db_jobs.job_get(job_id)
```
dadmobile, line 640: Same. Unnecessary?

transformerlab/routers/model.py
```diff
@@ -603,21 +628,30 @@ async def download_gguf_file_from_repo(model: str, filename: str, job_id: int |
...
+                await db_jobs.job_update_status(job_id, "FAILED", error_msg, experiment_id)
         return {"status": "error", "message": error_msg}
 
     # Validate the requested filename exists in the repository
     available_files = model_details.get("available_gguf_files", [])
     if filename not in available_files:
         error_msg = f"File '{filename}' not found in repository. Available files: {available_files}"
         if job_id:
-            await db_jobs.job_update_status(job_id, "FAILED", error_msg)
+            job = await db_jobs.job_get(job_id)
```
dadmobile, line 651: Again.  I will stop pointing these out as there are many.  IF for some reason this is necessary then we should factor this code and do something in the else condition I guess?

dadmobile (changes requested):

deep1401 (changes requested):
All functionality works except for the couple of comments on here and on the frontend PR

PR #484, opened 2025-07-26
Added IDE specific files to gitignore

michael-haroon:
In response to @dadmobile in PR https://github.com/transformerlab/transformerlab-app/pull/694

I added local IDE files to gitignore, namely .vscode, .cursor/, .cursorignore.

Let me know if it needs to be edited.


.gitignore
```diff
@@ -235,4 +235,12 @@ override.tf.json
 terraform.rc
 
 # Ignore .ssh folders
-**/.ssh/
\ No newline at end of file
+**/.ssh/
+
+# IDEs
+.vscode
```
dadmobile, line 241: Oh...we actually rely on a few files in .vscode to setup ruff and prettier for common linting settings.  I'm not sure if we should be doing that differently but we'd like to keep that folder included for now. :)

dadmobile (changes requested):

.gitignore
michael-haroon: Fixed it! Should be ready for merging.

dadmobile (approved):

PR #496, opened 2025-07-30
Add audio datasets support

deep1401:
Sample dataset to try: `raghib-blockstak/sample_audio_3`

### DIFF
--- a/transformerlab/routers/data.py
+++ b/transformerlab/routers/data.py
@@ -17,6 +17,8 @@ from datasets.data_files import EmptyDatasetError
 from transformerlab.shared.shared import slugify
 from transformerlab.shared import galleries
 from datasets.exceptions import DatasetNotFoundError
+import numpy as np
+import wave
 
 from transformerlab.db.datasets import (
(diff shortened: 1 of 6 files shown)


transformerlab/routers/data.py
```diff
@@ -218,8 +219,14 @@ async def dataset_preview(
 
 
 def serialize_row(row):
-    """Convert PIL Images in a row to base64 strings, preserving original structure."""
+    """Convert PIL Images and audio arrays in a row to base64 strings, preserving original structure."""
     if isinstance(row, dict):
+        # Check if this is an audio object
+        if "audio" in row and isinstance(row["audio"], dict):
+            audio_data = row["audio"]
+            if "array" in audio_data and "sampling_rate" in audio_data:
```
mina-parham, line 227: also same question for "array" and "sampling_rate", are they always there or should we handle cases where they might be missing?

mina-parham:
other than my comment I tested the PR and it works perfectly on my side

PR #510, opened 2025-08-08
Rewrite register process and kill worker processes properly

deep1401:


dadmobile:
Note: THis is tested and works, but we think we've fixed the root issue that this was dealing with. Deep and I are deciding on path forward based on further testing/investigation.

PR #512, opened 2025-08-12
Add/audio training

mina-parham:
Tested with model: 
- `unsloth/csm-1b` 
- `unsloth/orpheus-3b-0.1-ft`
and dataset: `MrDragonFox/Elise`

### DIFF
--- a/requirements-no-gpu-uv.txt
+++ b/requirements-no-gpu-uv.txt
@@ -574,7 +574,7 @@ tqdm==4.66.5
     #   peft
     #   sentence-transformers
     #   transformers
-transformerlab-inference==0.2.44
+transformerlab-inference==0.2.45
     # via -r requirements.in
 transformers==4.55.2
     # via
--- a/requirements-rocm-uv.txt
+++ b/requirements-rocm-uv.txt
@@ -577,7 +577,7 @@ tqdm==4.66.5
     #   peft
     #   sentence-transformers
     #   transformers
-transformerlab-inference==0.2.44
+transformerlab-inference==0.2.45
     # via -r requirements-rocm.in
 transformers==4.55.2
     # via
--- a/requirements-rocm.in
+++ b/requirements-rocm.in
@@ -31,7 +31,7 @@ markitdown[all]
 hf_xet
 macmon-python
 mcp[cli]
-transformerlab-inference==0.2.44
+transformerlab-inference==0.2.45
 diffusers==0.33.1
 pyrsmi
 controlnet_aux==0.0.10
--- a/requirements-uv.txt
+++ b/requirements-uv.txt
@@ -612,7 +612,7 @@ tqdm==4.66.5
     #   peft
     #   sentence-transformers
     #   transformers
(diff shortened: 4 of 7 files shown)


mina-parham:
> Just a note that this wont teach the model actual audio as the guide from Unsloth mentions this:
> 
> > The above is a simplification. In reality, to fine-tune Orpheus properly, you would need the audio tokens as part of the training labels. Orpheus’s pre-training likely involved converting audio to discrete tokens (via an audio codec) and training the model to predict those given the preceding text. For fine-tuning on new voice data, you would similarly need to obtain the audio tokens for each clip (using Orpheus’s audio codec). The Orpheus GitHub provides a script for data processing – it encodes audio into sequences of <custom_token_x> tokens.
> 
> However, Unsloth may abstract this away: if the model is a FastModel with an associated processor that knows how to handle audio, it might automatically encode the audio in the dataset to tokens. If not, you’d have to manually encode each audio clip to token IDs (using Orpheus’s codebook). This is an advanced step beyond this guide, but keep in mind that simply using text tokens won’t teach the model the actual audio – it needs to match the audio patterns.
> 
> So this wont work with Orpheus right, which is the most widely used model for fine-tuning right now.
> 
> May I suggest using this, which also uses unsloth but with snac decoding? https://huggingface.co/snorbyte/snorTTS-Indic-v0/blob/main/train_orpheus.py

thanks Deep! yeah I saw that, I'll use snac but  was planning to add it after I get some basic experiments with orpheus running, just to make sure the setup works first

transformerlab/plugins/unsloth_audio_trainer/index.json
```diff
@@ -0,0 +1,147 @@
...
+      "title": "Adam Epsilon",
+      "type": "number",
+      "minimum": 0.0,
+      "maximum": 1.0,
+      "default": 0.00000001
+    },
+    "adaptor_name": {
+      "title": "Adaptor Name",
+      "type": "string",
+      "required": true
```
dadmobile, line 119: This should probably have a default?

transformerlab/plugins/unsloth_audio_trainer/index.json
```diff
@@ -0,0 +1,147 @@
...
+  "version": "0.0.1",
+  "model_architectures": [
+    "CsmForConditionalGeneration"
+  ],
+  "supported_hardware_architectures": ["cuda"],
+  "git": "",
+  "url": "",
+  "files": ["main.py", "setup.sh"],
+  "setup-script": "setup.sh",
+  "training_template_format": "none",
```
dadmobile, line 16: I think we should mark this as experimental so we can merge it, and then mark public once we have a cuda loader?

dadmobile (changes requested):
Initial review...still running a train!

dadmobile (changes requested):
Can you add a simple info.md (see mlx_lora_trainer for an example). It should include anything weird you need to know to successfully run this (like the required format of the dataset, for example).

dadmobile (approved):

PR #514, opened 2025-08-15
Add/tts cuda

mina-parham:
tested for: 
1- `unsloth/csm-1b`
2- `unsloth/orpheus-3b-0.1-ft`


transformerlab/plugins/audio_server/main.py
```diff
@@ -0,0 +1,252 @@
...
+async def lifespan(app: FastAPI):
+    yield
+    # This function is called when the app shuts down
+    cleanup_at_exit()
+
+
+
+app = FastAPI(lifespan=lifespan)
+
+class UnslothAudioWorker(BaseModelWorker):
```
dadmobile, line 44: You called the UnslothAudioWorker although I don't know what is unsloth about it?  Or is there something I missed?
But then I think we use unsloth's libraries in the audio.py file instead of just huggingface transformers? 
I'm fine with whatever is easiest but maybe let's just name things right.

transformerlab/plugins/audio_server/audio.py
```diff
@@ -0,0 +1,129 @@
...
+            trimmed_row = row[:new_length]
+            trimmed_row = [t - 128266 for t in trimmed_row]  # Adjust to codec IDs
+            code_lists.append(trimmed_row)
+
+        # Run decoding
+        audio = [self.redistribute_codes(code_list) for code_list in code_lists]
+        return audio[0].squeeze().to(torch.float32).cpu().detach().numpy()
+
+    def redistribute_codes(self, code_list):
+            """Decode to waveform using SNAC"""
```
dadmobile, line 116: Wild...I don't knwo how you figured out to do this but maybe add a comment saying what you are doing approximately.  :)

transformerlab/plugins/audio_server/audio.py
```diff
@@ -0,0 +1,129 @@
...
+            max_seq_length=self.context_length,
+            dtype=None,
+            load_in_4bit=False,
+        )
+        FastModel.for_inference(self.model)
+        self.model = self.model.to(self.device)
+        # Tips for prompting [1]:
+        #  - Sampling parameters 'temperature' and 'top_p' work just like regular LLMs.
+        #  - 'repetition_penalty' >= 1.1 is required for stable generations.
+        #  - Increasing 'repetition_penalty' and/or 'temperature' makes the model speak faster.
```
dadmobile, line 71: Increasing temperature makes the model speak faster?!?

dadmobile (changes requested):

transformerlab/plugins/unsloth_text_to_speech_server/main.py
```diff
@@ -0,0 +1,244 @@
...
+
+        text = params.get("text", "")
+        model = params.get("model", None)
+        speed = params.get("speed", 1.0)
+        audio_format = params.get("audio_format", "wav")
+        sample_rate = params.get("sample_rate", 24000)
+        temperature = params.get("temperature", 0.0)
+        
+        
+        audio_dir = params.get("audio_dir", None)
```
dadmobile, line 104: Is there a reason to make this a parameter?  We could just avoid the warnings below altogether and put this in workspace?

dadmobile:
Interesting note...Fastchat wants to run the Orpheus model because it's architecture is Llama!  We might have to do more with `supports` in the future.

mina-parham:
> > > More importantly...Orpheus won't run because it's architecture doesn't match the supported architectures for the plugin!
> > 
> > 
> > Oh yeah I haven’t added the orpheus architecture yet because I think we first need to figure out a solution for supporting engine with audio models, probably something like checking the model tag also
> 
> OK I think we should create an issue to track? I think we're going to have to solve to some degree before we launch.

I'll create an issue for that :-)

transformerlab/plugins/unsloth_text_to_speech_server/index.json
```diff
@@ -0,0 +1,19 @@
...
+  "plugin-format": "python",
+  "type": "loader",
+  "version": "0.0.1",
+  "supports": [
+    "Text-to-Speech",
+    "Audio",
+    "experimental"
+  ],
+  "model_architectures": [
+    "CsmForConditionalGeneration"
```
dadmobile, line 14: And then for now I think you have to add Llama architecture here so Orpheus runs.  We'll create a separate issue to filter servers based on capabilities on the Foundation screen.

dadmobile (changes requested):

transformerlab/plugins/unsloth_text_to_speech_server/main.py
mina-parham: We get the experiment_dir based on experiment_id, so to do that here we need to add a function in plugin.py. I think we might get an error there, but I’ll see how to fix it here

PR #516, opened 2025-08-19
Fix Tool Calling

aahaanmaini:


transformerlab/fastchat_openai_api.py
```diff
@@ -415,6 +422,18 @@ async def get_gen_params(
     }
     if images is not None and len(images) > 0:
         gen_params["images"] = images
+    if tools is not None and len(tools) > 0:
+        gen_params["tools"] = tools
+        # Handle both regular tools (with function.name) and MCP tools (with direct name)
```
dadmobile, line 427: Let's chat here as well.  Why are there two kinds?  We should standardize on only MCP?

transformerlab/fastchat_openai_api.py
```diff
@@ -519,6 +537,49 @@ async def create_openapi_chat_completion(request: ChatCompletionRequest):
     if error_check_ret is not None:
         return error_check_ret
 
+    # NEW: Auto-load tools if not provided in request
```
dadmobile, line 540: Oh I don't think we want to do this here. It should be up to the caller to include tools on their request and this just processes!

transformerlab/fastchat_openai_api.py
```diff
@@ -1105,6 +1167,48 @@ async def create_chat_completion(request: APIChatCompletionRequest):
     if error_check_ret is not None:
         return error_check_ret
 
+        # NEW: Auto-load tools if not provided in request (for APIChatCompletionRequest too)
+    tools = None
```
dadmobile, line 1171: Same as above.

transformerlab/fastchat_openai_api.py
```diff
@@ -415,6 +418,18 @@ async def get_gen_params(
     }
     if images is not None and len(images) > 0:
         gen_params["images"] = images
+    if tools is not None and len(tools) > 0:
+        gen_params["tools"] = tools
+        # Extract tool names for logging (MCP tools only)
```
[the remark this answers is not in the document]
aahaanmaini: Primarily added for debugging, removed!

dadmobile (approved):

PR #525, opened 2025-08-27
Add/audio cloning mlx

mina-parham:


transformerlab/plugins/mlx_audio_server/setup.sh
```diff
@@ -1,6 +1,4 @@
 #!/usr/bin/env bash
 #uv pip install mlx==0.24.0 # Using this version to work around bug: https://github.com/Blaizzy/mlx-audio/issues/207
-uv pip install "mlx-lm==0.24.0"
```
dadmobile, line 3: You can remove this because mlx-audio automatically installs mlx-lm?  And before you were just including so you could peg the version of mlx-lm?

transformerlab/plugins/mlx_audio_server/main.py
```diff
@@ -78,6 +78,10 @@ async def generate(self, params):
         sample_rate = params.get("sample_rate", 24000)
         temperature = params.get("temperature", 0.0)
         stream = params.get("stream", False)
+        # Audio cloning parameters
+        audio_path = params.get("audio_path", None)  # Reference audio file for voice cloning
+        ref_text = params.get("ref_text", None)    # Text spoken in reference audio (optional)
+        stt_model = params.get("stt_model", "mlx-community/whisper-large-v3-turbo")  # STT model for transcribing ref_audio
```
dadmobile, line 84: Oooooh right. So what happens if you don't have this model? It automatically downloads and you sit there waiting?

mina-parham: yes it seems mlx-audio handles this part, but I haven’t tried it yet. If it takes too long, I’ll disable it and instead just add a field in the ui where the user can enter the transcription for ref_audio

dadmobile (changes requested):
This didn't work for me!  I tried with the same of my voice that worked for CUDA+Orpheus and it did nothing on MLX and Kokoro.  Is that the model I should be using? Or is there a different one?

PR #526, opened 2025-08-27
remove pinned mlx version and spacy

mina-parham:


deep1401 (approved):
Tested and this works

PR #527, opened 2025-08-28
Fix/worker health race condition

mina-parham:
solves #743

https://github.com/transformerlab/transformerlab-app/issues/743

These changes address a race condition between the `/refresh_all_workers` and `/list_models` endpoints. Previously, after triggering a refresh of all workers, the code would immediately request the list of available models. However, there was no guarantee that all workers had finished re-registering, which could result in /list_models returning an empty or incomplete list.

To solve this, we now poll /list_models after calling /refresh_all_workers, waiting until a non-empty model list is returned or a timeout is reached. This ensures that the API only returns the list of models after the refresh operation is complete

### DIFF
--- a/transformerlab/fastchat_openai_api.py
+++ b/transformerlab/fastchat_openai_api.py
@@ -466,15 +466,24 @@ async def show_available_models():
     controller_address = app_settings.controller_address
     async with httpx.AsyncClient() as client:
         await client.post(controller_address + "/refresh_all_workers")
-        await asyncio.sleep(0.5)  # Allow time for workers to re-register after refresh
-        ret = await client.post(controller_address + "/list_models")
-    models = ret.json()["models"]
-    models.sort()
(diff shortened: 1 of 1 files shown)


transformerlab/fastchat_openai_api.py
```diff
@@ -466,14 +467,24 @@ async def show_available_models():
...
+        elapsed = 0.0
+        models = []
+        while elapsed < timeout:
+            ret = await client.post(controller_address + "/list_models")
+            models = ret.json().get("models", [])
+            if models:
+                break
+            await asyncio.sleep(poll_interval)
+            elapsed += poll_interval
+        models.sort()
```
dadmobile, line 482: How come you moved this into the `with` block? Any chance this can fail to return something in an error state or something?

dadmobile (approved):

PR #528, opened 2025-08-28
Add/voice audio

mina-parham:


transformerlab/fastchat_openai_api.py
```diff
@@ -518,6 +518,13 @@ async def create_audio_tts(request: AudioRequest):
         "speed": request.speed,
         "audio_path": request.audio_path,
     }
+    
+    # Add voice parameter if provided
+    if request.voice:
```
dadmobile, line 523: Possibly you need to check if voice is not empty here?  I'm not sure if the error is from voice being empty or in the way the app handles output (see other comment).

dadmobile (changes requested):
With the latest change you can select language and not select a voice and it throws an error.  The first time i did it I got an alert with the model name in it.  Every time since I just get this:
image

 Looking at transformerlab.log I just see message saying `Audio Successfully generated` but actually all it generated was a json file.  The audio file is not there!

PR #529, opened 2025-08-29
Fix python environment and internal errors of yourbench plugin

deep1401:
This adds a special condition to create the plugin venv at python 3.12 for yourbench.

Fixes transformerlab/transformerlab-app#750

### DIFF
--- a/transformerlab/routers/plugins.py
+++ b/transformerlab/routers/plugins.py
@@ -200,9 +200,16 @@ async def install_plugin(plugin_id: str):
         print("Creating virtual environment for plugin...")
         await log_file.write(f"## Creating virtual environment for {plugin_id}...\n")
 
-        proc = await asyncio.create_subprocess_exec(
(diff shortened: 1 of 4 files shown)


transformerlab/routers/plugins.py
```diff
@@ -200,9 +200,16 @@ async def install_plugin(plugin_id: str):
         print("Creating virtual environment for plugin...")
         await log_file.write(f"## Creating virtual environment for {plugin_id}...\n")
 
-        proc = await asyncio.create_subprocess_exec(
-            "uv", "venv", venv_path, "--python", "3.11", cwd=new_directory, stdout=log_file, stderr=log_file
-        )
+        if "yourbench" in plugin_id:
```
dadmobile, line 203: Can you add a comment here explaining why we did this and maybe even linking to the github?

dadmobile (changes requested):

dadmobile (approved):

PR #530, opened 2025-08-29
Make new uv venv environments with clear flag

deep1401:


transformerlab/routers/plugins.py
```diff
@@ -203,11 +203,27 @@ async def install_plugin(plugin_id: str):
         if "yourbench" in plugin_id:
             await log_file.write("## Setting up python 3.12 for yourbench...\n")
             proc = await asyncio.create_subprocess_exec(
-                "uv", "venv", venv_path, "--python", "3.12", cwd=new_directory, stdout=log_file, stderr=log_file
+                "uv",
+                "venv",
+                venv_path,
+                "--python",
+                "3.12",
+                "--clear",
```
dadmobile, line 211: Dumb question but do we want to clear the environment every time? That probably makes sense but I didn't think we were doing that now?

PR #541, opened 2025-09-04
Fix job_get and remove job_get_status and job_get_error_msg

deep1401:
Fixes transformerlab/transformerlab-app#754

The `experiment_id=None` cases are for the things where we try to get the experiment_id from the actual job instead of having it already


dadmobile (changes requested):
I'm not sure I understand why you added experiment_id back to job_get?

dadmobile:
Yeah I think at this point we are all in on not needing experiment ID for anything you can get from a standard job_get.  So I think it's better to just get rid of that parameter.  Sorry for the pain!

deep1401:
> Yeah I think at this point we are all in on not needing experiment ID for anything you can get from a standard job_get. So I think it's better to just get rid of that parameter. Sorry for the pain!

Okay fixed now, job_get does not take experiment_id anywhere now

transformerlab/shared/shared.py
```diff
@@ -377,7 +377,7 @@ async def run_job(job_id: str, job_config, experiment_name: str = "default", job
             # Only set to COMPLETE if not already FAILED
             job = await db_jobs.job_get(job_id)
             experiment_id = job["experiment_id"]
-            current_status = await db_jobs.job_get_status(job_id, experiment_id)
+            current_status = job["status"]
```
dadmobile, line 380: It's probably a good idea to use .get("status") for these accesses?  I think we always set status, but somehow this always bites me.  :)

dadmobile (changes requested):

transformerlab/shared/shared.py
deep1401: Made the changes, ready to review again!

PR #553, opened 2025-09-08
Use os.kill() for worker exit

deep1401:
Fixes transformerlab/transformerlab-app#696

It still terminates with a SIGTERM (-15) signal but the lifespan is correctly executed now for termination


dadmobile:
OK this looks like it's doing the right thing now. I still don't think we were always getting a -15 but I guess we can deal with that (perhaps we weren't SIGTERMing before?).

PR #563, opened 2025-09-15
Add S3 Bucket Mounting with Mountpoint for Multitenant mode

deep1401:


deep1401:
> Do we still need to do the install.sh bit if you added the user service part?

We still need to install mountpoint right. But if you like we could just make a separate install.sh for multitenant instead of adding things to our standard one

dadmobile (dismissed):

api.py
```diff
@@ -65,10 +65,12 @@
 from transformerlab.routers.experiment import jobs
 from transformerlab.shared import shared
 from transformerlab.shared import galleries
-from transformerlab.shared.constants import WORKSPACE_DIR
+from lab.dirs_workspace import get_workspace_dir
```
dadmobile, line 68: Any reason you chose lab.dirs_workspace vs. lab.dirs?  I wonder if the latter is cleaner and maybe we will have to use it anyways in case there is something else to import?

deep1401: I just did everything in dirs_workspace, we could essentially move everything to dirs.py and just get rid of dirs_workspace.py

deep1401: Deleted dirs_workspace now, everything comes from lab.dirs

transformerlab/shared/shared.py
```diff
@@ -761,7 +763,7 @@ def on_job_complete():
 
                 # Create a unique adaptor directory for this run
                 run_adaptor_dir = os.path.join(
-                    WORKSPACE_DIR, "adaptors", secure_filename(model_name), f"{adaptor_name}_sweep_{i + 1}"
+                    workspace_dir, "adaptors", secure_filename(model_name), f"{adaptor_name}_sweep_{i + 1}"
```
dadmobile, line 766: Will make a note to do a pass over directories in API later on.

PR #566, opened 2025-09-16
Fix plugin lookup, refactor library enumeration, and update pytest

mashcroft3:
### Summary

This PR introduces a bug fix for plugin database lookups, a performance and reliability refactor for Python package enumeration, and an enhancement to the pytest configuration.

---

### Changes

* **`pytest.ini`**:
    * Added `norecursedirs` to the `[pytest]` configuration. This prevents the test collector from scanning virtual environments (`*/venv/*`, `*/site-packages/*`) and temporary directories (`test/tmp`), improving performance and preventing the execution of third-party tests.

* **`transformerlab/db/db.py`**:
    * **BUG FIX**: The `save_plugin` function was modified to query for plugins using `select(Plugin).where(Plugin.name == name)` instead of `session.get(Plugin, name)`.
    * **Reason**: The previous implementation incorrectly assumed the `name` field was the primary key. This change corrects the lookup logic to query against the unique `name` column, resolving an issue where existing plugins could not be found and updated.

* **`transformerlab/routers/serverinfo.py`**:
    * **REFACTOR**: The `get_python_library_versions` endpoint was updated to use `importlib.metadata` to enumerate installed packages.
    * **Reason**: This removes the need for a `subprocess` call to `pip`, which improves endpoint performance and security. The previous `pip` method is retained as a fallback for increased resilience.

### DIFF
--- a/transformerlab/db/db.py
+++ b/transformerlab/db/db.py
@@ -631,12 +631,14 @@ async def get_plugin(slug: str):
 
 async def save_plugin(name: str, type: str):
     async with async_session() as session:
-        plugin = await session.get(Plugin, name)
-        if plugin:
-            plugin.type = type
-        else:
+        # Plugin primary key is integer id, so we must query by unique name
+        result = await session.execute(select(Plugin).where(Plugin.name == name))
+        plugin = result.scalar_one_or_none()
+        if plugin is None:
             plugin = Plugin(name=name, type=type)
             session.add(plugin)
+        else:
+            plugin.type = type
         await session.commit()
     return
 
--- a/transformerlab/routers/serverinfo.py
+++ b/transformerlab/routers/serverinfo.py
@@ -265,12 +265,39 @@ async def get_computer_information():
 
 @router.get("/python_libraries")
 async def get_python_library_versions():
-    # get the list of installed python packages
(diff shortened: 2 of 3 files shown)


transformerlab/db/db.py
```diff
@@ -631,12 +631,14 @@ async def get_plugin(slug: str):
 
 async def save_plugin(name: str, type: str):
```
deep1401, line 632: I think this is a leftover function, we dont use this one anymore or the associated parent function, I think?

transformerlab/routers/serverinfo.py
```diff
@@ -265,12 +265,39 @@ async def get_computer_information():
 
 @router.get("/python_libraries")
```
deep1401, line 266: Tried this change and it worked for me

deep1401 (changes requested):
All changes work for me but I think the plugin one might be unnecessary as that function doesn't get used

dadmobile:
This looks good to me.  The plugin change is unnecessary but I made a separate issue to remove the tests/db code for plugins.

dadmobile (approved):

deep1401 (approved):

PR #573, opened 2025-09-19
Use env var for TFL_SOURCE_CODE_DIR

mina-parham:
the source code path is being set based on `os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))` to compute paths based on the file's location. this would create problem in sdk. to fix this I handle the source code path with env variable


api.py
```diff
@@ -73,14 +73,18 @@
 
 load_dotenv()
 
+api_source_dir = os.path.dirname(os.path.abspath(__file__))
```
dadmobile, line 76: Can you put a comment on this? Say something explaining that we are using the location of this file to track the directory that the API code is being served from.

dadmobile (changes requested):

PR #589, opened 2025-09-25
Update tlab inference to 0.2.49, sdk to 0.0.7 and update server plugins

deep1401:


constraints.txt
```diff
@@ -1 +0,0 @@
-torch==2.6.0
```
dadmobile, line 1: ???

PR #592, opened 2025-09-26
Move models db to filesystem

deep1401:
Depends on transformerlab/transformerlab-sdk#14
To only be merged after that, tests will fail until then


dadmobile:
OK this touches so much with import and different kinds of models etc.  I've tested what I can and it looks good and the tests pass.  So I think we merge and since we have so much new stuff we're going to have to do a huge passover of the app anyways.

dadmobile:
But...last thing before approving, I'd love to release the SDK v 0.0.12 just to verify that everything is running well here. (it is causing test failures here)

PR #593, opened 2025-09-28
Fix/experiment id name

mina-parham:


api.py
```diff
@@ -87,6 +91,106 @@
...
+                    print("No experiments in DB to migrate.")
+                    return
+
+                for exp in experiments:
+                    print(f"Migrating experiment: {exp['name']}")
+                    
+                    # Create SDK Experiment
+                    experiment = Experiment.create(exp['name'])
+                    # TODO: which datetime to use here?
+                    experiment._update_json_data_field(key="id", value=exp["name"])
```
dadmobile, line 134: Let's store the old experiment_id somewhere just in case?  Put in a field called legacy_id or db_id or something?

api.py
```diff
@@ -87,6 +91,106 @@
...
+
+                    print(f"Created experiment directory: {experiment.get_dir()}")
+
+async def migrate_db_to_filesystem():
+    """Migrate data from DB to filesystem if not already migrated."""
+    
+    try: 
+        # Check if experiments migration is needed
+        # We need migration if there are experiment directories WITHOUT index.json
+        experiments_need_migration = True
```
dadmobile, line 147: Logic is weird here.  You start with experiments_need_migration as True and then only ever update it to True.  So I think it's always true.

api.py
```diff
@@ -87,6 +91,106 @@
...
+                exp_path = os.path.join(lab_dirs.EXPERIMENTS_DIR, exp_dir)
+                if os.path.isdir(exp_path):
+                    index_file = os.path.join(exp_path, "index.json")
+                    if not os.path.exists(index_file):
+                        experiments_need_migration = True
+                        break
+        
+        # Check if jobs migration is needed
+        # We need migration if there are job directories WITHOUT index.json
+        jobs_need_migration = False
```
dadmobile, line 159: Opposite of experiments...you start False adn only update to be False.  So I think this is always False.

api.py
```diff
@@ -87,6 +91,106 @@
...
+            for job_dir in os.listdir(lab_dirs.JOBS_DIR):
+                job_path = os.path.join(lab_dirs.JOBS_DIR, job_dir)
+                if os.path.isdir(job_path):
+                    job_index = os.path.join(job_path, "index.json")
+                    if not os.path.exists(job_index):
+                        jobs_need_migration = False
+                        break
+        
+        # If neither needs migration, skip entirely
+        if not experiments_need_migration and not jobs_need_migration:
```
dadmobile, line 170: I think you can assume there is NO situation where you do jobs but not experiments?  How would you get into that situation? Either migrate or don't.

api.py
```diff
@@ -87,6 +91,106 @@
...
+            print("Migration not needed - all experiment and job directories have index.json files.")
```
dadmobile, line 171: Probably don't say anything if migration is done.  This is going to be the "normal" situation going forward so this message will always show on startup.

api.py
```diff
@@ -87,6 +91,106 @@
...
+                        job_obj._update_json_data_field(key="id", value=job["id"])
+                        job_obj._update_json_data_field(key="experiment_id", value=exp["name"])  # Use name instead of numeric ID
+                        job_obj._update_json_data_field(key="job_data", value=job["job_data"] if isinstance(job["job_data"], dict) else {})
+                        job_obj._update_json_data_field(key="status", value=job["status"])
+                        job_obj._update_json_data_field(key="type", value=job["type"])
+                        job_obj._update_json_data_field(key="progress", value=job["progress"])
+                        job_obj._update_json_data_field(key="created_at", value=job["created_at"])
+                        job_obj._update_json_data_field(key="updated_at", value=job["updated_at"])
+
+                        print(f"Created job directory: {job_obj.get_dir()}")
```
dadmobile, line 118: So not related but we should probably do in this PR...any jobs created in the last month are going to be in the wrong place (in experiments/jobs).  Could we add a thing that if you're moving a job it checks to see if it's directory is in the wrong place and moves it to the jobs directory?  Pain in the butt but it will fix a lot and simplify for us.

transformerlab/db/db.py
```diff
@@ -358,235 +362,87 @@ async def export_job_create(experiment_id, job_data_json):
...
-    if not isinstance(config, dict):
-        raise ValueError("Config must be a dictionary")
-    async with async_session() as session:
-        experiment = models.Experiment(name=name, config=config)
-        session.add(experiment)
-        await session.commit()
-        await session.refresh(experiment)
-        return experiment.id
-
+    experiments = []
```
dadmobile, line 365: I think I said this elsewhere: We can't do any of this in the DB code.  WE should change this to call the SDK and let the SDK do the internal work (not call _update_json_data_field ideally).  If there's stuff the SDK should be doing to support then we can add there.

dadmobile: I realize this might add overhead because maybe there's 5 places in the code that call a DB function.  This is why I made the experiment_service file so I could centralize and then route appropriately.  But maybe we just change DB calls to SDK calls and avoid...not sure.  Let's discuss.

dadmobile (changes requested):

api.py
mina-parham: sorry,  wanted to skip the migration on my local and accidentally pushed the changes

PR #594, opened 2025-09-29
Fix workflow test fail

mina-parham:
The problem was that in the tests, both the home and workspace directories were set to the same place. Because of that, the old database path and the new one ended up being identical. So when the migration code tried to copy the database, it was actually copying it onto itself


test/api/conftest.py
```diff
@@ -2,10 +2,13 @@
 from fastapi.testclient import TestClient
 import os
 
+# Create test directories before setting environment variables
+os.makedirs("test/tmp/workspace", exist_ok=True)
```
dadmobile, line 6: Nitpick: It seems like you shouldn't have to do this.  There's even a test to test that if you delete this directory it creates a new one?

dadmobile (approved):

PR #596, opened 2025-10-02
Move tasks db to filesystem

deep1401:
Depends on transformerlab/transformerlab-sdk#19

### DIFF
--- a/transformerlab/routers/tasks.py
+++ b/transformerlab/routers/tasks.py
@@ -3,7 +3,6 @@ import json
 from fastapi import APIRouter, Body
 from werkzeug.utils import secure_filename
 
-import transformerlab.db.db as db
 from lab import Dataset
 from transformerlab.db.jobs import job_create
 from transformerlab.models import model_helper


transformerlab/services/tasks_service.py
```diff
@@ -0,0 +1,108 @@
...
+Tasks service that uses the filesystem instead of the database.
+This replaces the database-based task operations with filesystem-based ones.
+"""
+
+import uuid
+from typing import List, Dict, Any, Optional
+from lab.task import Task as TaskService
+
+
+class TasksService:
```
dadmobile, line 11: Is there a reason to make these async?

transformerlab/services/tasks_service.py
```diff
@@ -0,0 +1,108 @@
...
+    """Service for managing tasks using filesystem storage"""
+    
+    def __init__(self):
+        self.task_service = TaskService
```
dadmobile, line 15: Not asking for a change but I don't totally understand what the benefit of wrapping this in a class and making a singleton is?

dadmobile (changes requested):

transformerlab/services/tasks_service.py
deep1401: There is no reason, I just wanted to structure these better so I started with the tasks one and thought I'd keep adding similar things for the other services. I can remove that singleton from __init__ if that is more preferable

dadmobile (approved):

PR #598, opened 2025-10-03
added /model/logout_from_huggingface API for clearing the hftoken

salsina:
To address issue #703 on the frontend app (Add option to update Hugging Face credentials in the UI), the backend should remove the cached Hugging Face token. This requires logging out from Hugging Face before setting the new token.


deep1401:
Thanks for your contribution! :)

PR #600, opened 2025-10-06
Change all dirs to functions

deep1401:
Depends on transformerlab/transformerlab-sdk#22

Tests will fail before merging the sdk PR

### DIFF
--- a/api.py
+++ b/api.py
@@ -121,9 +121,6 @@ async def migrate():
             await tasks.convert_all_to_tasks(exp["id"])
 
 
-
-
-
 async def run_over_and_over():
     """Every three seconds, check for new jobs to run."""
     while True:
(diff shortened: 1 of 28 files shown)


api.py
```diff
@@ -364,7 +364,8 @@ async def server_worker_start(
 
     print("Loading plugin loader instead of default worker")
 
-    with open(lab_dirs.GLOBAL_LOG_PATH, "a") as global_log:
+    from lab.dirs import get_global_log_path
```
dadmobile, line 367: Out of curiosity, why the late import on these?

requirements-no-gpu-uv.txt
```diff
@@ -548,7 +548,7 @@ tqdm==4.66.5
     #   peft
     #   sentence-transformers
     #   transformers
-transformerlab==0.0.13
```
dadmobile, line 551: Latest version will be 0.0.16.  Sorry...maybe we just remove the version number for now until we get to a stable update state or something?

dadmobile (changes requested):
Sorry about screwing up the version number.  Other than that we can approve. I'm curious to know what the late imports are about!

api.py
deep1401: I had a task failing in model.py where the late import was the only thing that worked. So I did late imports wherever possible so the context is set correctly when importing

PR #679, opened 2025-11-13
[WIP] Added Support for importing model from lm studio

ParamThakkar123:
Fixes https://github.com/transformerlab/transformerlab-app/issues/160

Should I also go ahead and add a lmstudio server in plugins? Because lmstudio runs on localhost:1234


dadmobile:
Is this still WIP?  

1. There are some CodeQL issues I can help you fix...mostly just see how we use secure_filename elsewhere and copy.
2. If you want to try adding the LM Studio plugin let's do that in a separate PR after we migrate API over to the app. Can you remove that plugin code for now (or backup and we can discuss in a new PR).

transformerlab/models/model_helper.py
```diff
@@ -65,6 +66,8 @@ def get_model_by_source_id(model_source: str, model_source_id: str):
                 return ollamamodel.OllamaModel(model_source_id)
             case "huggingface":
                 return huggingfacemodel.HuggingFaceModel(model_source_id)
+            case "lmstudio":
```
dadmobile, line 69: I think you'll also have to add a case for "lmstudio" to the `list_models_from_source` function so they show up on the import modal.

transformerlab/models/lmstudiomodel.py
```diff
@@ -0,0 +1,110 @@
...
+                "name": self.name,
+                "model_filename": output_filename,
+                "description": f"LM Studio model {self.source_id_or_path}",
+                "source": self.source,
+                "architecture": json_data["architecture"],
+            },
+        }
+
+        model_info_file = os.path.join(output_path, "index.json")
+        with open(model_info_file, "w") as f:
```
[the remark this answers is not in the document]
dadmobile: You can probably use this pattern anywhere that it is complaining about user-assigned values for paths.

transformerlab/plugins/lmstudio_server/index.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Google LM Studio Server",
```
dadmobile, line 2: Let's pull this for now and discuss if we actually want to add a server or not.

dadmobile (changes requested):

ParamThakkar123:
Okay @dadmobile Probably what we can do is close this PR for now and keep this as a backup and later discuss about it separately and I will open it once again

deep1401:
@ParamThakkar123 @dadmobile Closing this, not sure what we decided to do with this one but if need be you can reopen on transformerlab-app repo

PR #687, opened 2025-11-15
Added server logging for displaying feedback while loading models

ParamThakkar123:
I am getting these errors quite frequently from the API after I fetched the latest version from main

image

### DIFF
--- a/api.py
+++ b/api.py
@@ -7,6 +7,7 @@ import argparse
 import asyncio
 import logging
 import traceback
+import time
 
 import json
 import signal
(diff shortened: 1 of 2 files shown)


ParamThakkar123:
> I am getting these errors quite frequently from the API after I fetched the latest version from main
> 
> image

This one got fixed after plugin reinstallation

ParamThakkar123:
Depends on https://github.com/transformerlab/transformerlab-app/pull/882

deep1401:
> I am getting these errors quite frequently from the API after I fetched the latest version from main
> 
> image

This is probably caused by an older version of the plugin. The error will go away when you update your plugin

deep1401 (changes requested):
tail field in job data stored things correctly for me, but I think the initial startup of the subprocess isthe one that takes time and there is no output in console then. Which is why you'll only see `Starting model...` for most plugins while only some whose actual process also takes time, those will show these things. 

Maybe @dadmobile can help us resolve but I think this wont solve the problem right? 

Also side note, we wouldn't need to add a new server logs route in API right? 
I thought you were just reading from job data directly on the frontend already


Just for reference this is what `tail` had in job data:
```
{
  "tail": [
    "2025-11-17 07:55:03 | INFO | model_worker | Loading the model ['Qwen3-0.6B'] on workerb617e619, worker type: MLX worker...",
    "2025-11-17 07:55:03 | INFO | model_worker | Model architecture: Qwen3ForCausalLM",
    "2025-11-17 07:55:04 | ERROR | stderr | \rFetching 7 files:   0%|          | 0/7 [00:00<?, ?it/s]",
    "2025-11-17 07:55:04 | ERROR | stderr | \rFetching 7 files: 100%|██████████| 7/7 [00:00<00:00, 30551.64it/s]",
    "2025-11-17 07:55:04 | ERROR | stderr |",
    "2025-11-17 07:55:05 | INFO | stdout | Context length:  40960"
  ]
}
```

ParamThakkar123:
> Changes

I updated the code to use the JobData and removed a separate serverlog route

deep1401:
To be ported into https://github.com/transformerlab/transformerlab-app/pull/882

ParamThakkar123:
@deep1401 I already added this code to the linked pull request on the app. We can probably close this :)

deep1401:
Code has been ported to App so closing this

